PR #52, opened 2011-02-25
Options to toggle 'Bookmarks' and 'Press ? for help'

camthompson:
Adds options (NERDTreeBmarksLabel and NERDTreePressHelp, respectively) to hide the 'Bookmarks' label and 'Press ? for help' text inside the NERD tree window.

### DIFF
--- a/plugin/NERD_tree.vim
+++ b/plugin/NERD_tree.vim
@@ -2668,6 +2668,7 @@ function! s:initNerdTreeInPlace(dir)
         setlocal nu
     else
         setlocal nonu
+        setlocal nornu
     endif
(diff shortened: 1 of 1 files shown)


uri:
Is there any way to disable the press `?` for help? I frequently use Vim's default behaviour for `?`.

scrooloose:
@uri In your `~/.vimrc` you can set it to F1 with  `let NERDTreeMapHelp='<f1>'` or `let NERDTreeMapHelp=''` to disable it altogether.

uri:
Great thanks @scrooloose!

PR #72, opened 2011-06-15
"Deep Find"

benjamingeiger:
In my Rails workflow, I open vim from the top level directory of the source tree. I want to keep the root of the tree there at all times. But I also want to use :NERDTreeFind so I don't have to dig down four or five levels to find a file.

So, I've implemented a feature I call "deep find", which does precisely that. If the current file is a descendant of the current directory, the tree opens rooted at the current directory but with the current file highlighted. (If the current file is not a descendant of the current directory, the command behaves as before.)

I have also included a configuration variable ("NERDTreeDeepFind") to enable or disable the feature; it defaults to disabled.


nkcfan:
Could you add this feature? An option to NERDTreeFind seems quite good.

pinggit:
how to get this deep find feature work? it seems the lastest still does not have it.

PR #102, opened 2011-12-16
Prevent nerdtree from splitting other window, when that is just empty (&buftype=='nofile')

daniel-kullmann:
I often have the situation where I close all existing buffers (except the NERDTree buffer), and then start opening files from NERDTree again.

When doing this, NERDTree opens the files in a new split window, i.e. I have NERDTree at the left side, and on the right side two windows on top of each other, an emtpy buffer and the file I wanted to open.

What I want to happen is that NERDTree opens the file in the right window without splitting it.
This commit fixes this problem, although I'm not sure whether this is the best way to accomplish
this..

### DIFF
--- a/plugin/NERD_tree.vim
+++ b/plugin/NERD_tree.vim
@@ -3068,7 +3068,7 @@ function! s:firstUsableWindow()
     let i = 1
     while i <= winnr("$")
         let bnum = winbufnr(i)
-        if bnum != -1 && getbufvar(bnum, '&buftype') ==# ''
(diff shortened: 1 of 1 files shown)


mhinz:
@scrooloose Apparently a split will happen for all windows that are normal buffers (`buftype=`). My guess is this happens because of plugin compatability, so, let's say, TagBar isn't closed by accident.

Now there are scratch buffers (`buftype=nofile`), that are "throwaway" buffers. They're kind of placeholders. In these cases you woudn't want a split, but the to be opened file to inherit the window showing the scratch buffer.

Is there anything we can do about, without changing other plugins?

PR #166, opened 2012-05-23
ability to "lynx-like" motion

kucaahbe:
behavior: [only] on directory node right arrow(<Right>) opens it, left arrow(<Left>) closes without affecting child nodes state. Came from midnight commander's "Lynx-like motion" option.

P.S.
I'm not sure is this done right, maybe it make sense to add to README (or wiki) "howto" about extending nerdtree to have this functionality. So, what do you think?


PhilRunninger:
I'm closing this issue because of its age and the fact that it conflicts with the current source code. Also this functionality would be better if done in either the vimrc or a NERDTree plugin using the NERDTree API.

Issue #154, opened 2012-04-16
With NERDTreeShowHidden set, special folders . and .. should not be displayed


mihaifm:
I've just set the NERDTreeShowHidden option and the screen became cluttered with . and .. folders on each directory. 

I just want the .gitignore, .vimrc and other files displayed. `.` and `..` should be hidden even with this option set. There's no use for them on screen.

mihaifm:
This doesn't replicate any more, I'm closing it since it's rather old.

PR #207, opened 2012-11-11
Ignore .. and . special dirs regardless of the NERDTreeShowHidden option

mihaifm:
This fixes [#154](https://github.com/scrooloose/nerdtree/issues/154)


PhilRunninger:
@mihaifm, Is this issue still happening to you? I can't reproduce it on either Mac or Windows.

mihaifm:
I can't replicate it any longer as well. 
It's an old item, it must have been fixed in the meantime.

PR #313, opened 2014-02-16
Horizontal split - relative to the previous window

paprykarz:
This patch causes splitting recently visited window with `i` key instead of the window adhering to the left or right - it gives you more control of where to place a new window.

### DIFF
--- a/lib/nerdtree/opener.vim
+++ b/lib/nerdtree/opener.vim
@@ -92,7 +92,7 @@ function! s:Opener._newSplit()
     " splitbelow and splitright IF the explorer is the only window.
     "
     let there= g:NERDTreeWinPos ==# "left" ? "wincmd h" : "wincmd l"
-    let back = g:NERDTreeWinPos ==# "left" ? "wincmd l" : "wincmd h"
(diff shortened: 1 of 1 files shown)


dragonxlwang:
Though the above issue is hard to resolve, would it be possible to have @paprykarz's pr as a temp solution and add as an option?

Reason is that currently nerd tree vsplit behaves to split at the previous window but nerdtree split doesn't. 

I usually have 3 column editors arranged so that if I jump from 2nd/3rd column to nerdtree and do a split it's a bit annoying to see it spawns at the 1st column;

aantonello:
The way I see this is that Vim users are used to this behavior. I mean, who does like splitted windows to open below the current one will set `splitbelow` and who does like splitted windows to open at right will set `splitright`.

I have seen others Project/Organizer plugins solving this issue in a simple way, using `wincmd p` to go back to the previous window and issuing the `sp <file>` or `vsp <file>` from there. I missed this behavior in NERDTree when I installed it. And changed my local copy to follow this rule. A new option is good for new Vim users. The default behavior should be the one that old Vim users are already used to.

PR #318, opened 2014-02-27
Add "search filter" functionality

gorcha:
For your consideration - added a new function to filter the file list based on a regular expression.
Have added with key mapping "G", it turns off the filter when it's enabled.


PhilRunninger:
@gorcha , this would be a great candidate for a NERDTree plugin. I don't think it should be in the core NERDTree.

lifecrisis:
I suggest closing this PR.  @PhilRunninger, your call.

PR #364, opened 2014-07-26
Don't ignore autocmd events when close tree

cohama:
I (and maybe some users) have the following setting

```vim
autocmd WinEnter * setlocal cursorline
autocmd WinLeave * setlocal nocursorline
```

This sets `cursorline` only on the current window.

But, when the NERDTree window closed, `cursorline` appears on not-current window because WinEnter/WinLeave events are ignored.

In addition, same issue will occur when using [Powerline](https://github.com/Lokaltog/powerline).
`Powerline` handles 'statusline' in the same way.

This PR makes tree window fire the autocmd events when closed.

### DIFF
--- a/autoload/nerdtree.vim
+++ b/autoload/nerdtree.vim
@@ -135,7 +135,7 @@ function! nerdtree#closeTree()
 
         call nerdtree#exec(nerdtree#getTreeWinNum() . " wincmd w")
         close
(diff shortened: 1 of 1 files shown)


PhilRunninger:
@cohama, NERDTree has been refactored since your PR, and this doesn't appear to be happening any more. Is it OK to close this?

PR #380, opened 2014-09-01
New configuration option: NERDTreeActivateFileMode

letorbi:
Hi,

_The original commits are not valid anymore, see my second comment for the current decription_

~~I've made some small changes to make NERDTree open every file in a new window. This way every file will stay opened until closed explicitly. If a file has already been opened in a window, this window will be brought to front instead of creating a new one. I'm using NERDTreeAlwaysSplit for about two years now and really like it, because it creates some kind of "tab-like" feeling.~~ 

~~As you can see in commit e77accd the changes in the code are really small and the always split functionality is disabled by default. In the two years of using my patch I never experienced any bugs or side-effects caused by my changes.~~ 

Hope hearing from you :)

Baibai
Torben

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -99,7 +99,7 @@ endfunction
 "FUNCTION: s:activateFileNode() {{{1
 "handle the user activating a tree node
 function! s:activateFileNode(node)
-    call a:node.activate({'reuse': 'all', 'where': 'p'})
+    call a:node.activate({'reuse': 'all', 'where': g:NERDTreeActivateFileMode})
 endfunction
 
 "FUNCTION: s:activateBookmark() {{{1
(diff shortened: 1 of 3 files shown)


letorbi:
I have improved the original functionality: `NERDTreeAlwaysSplit` has been replaced by `NERDTreeActivateFileMode`, which allows to modify the behaviour of the `activateFileNode` function. The option is now more generic (it's possible to set vertical and horizontal splits as well as tabs and of cause the default behaviour) and applies it's changes a bit more "upstream" than the original commit.

I hope this fits your requirements and that my commit gets merged into the main repo. If not, please tell me the reasons why, so I can fix them.

PhilRunninger:
I recommend rejecting this PR. Adding a new option to circumvent learning the keyboard shortcuts isn't something I'd be willing to support. @lifecrisis, what do you say?

lifecrisis:
Agreed.  Thanks for the suggestion, @letorbi.

PR #404, opened 2014-11-21
Updated nerdtree expand to use resolve to handle symlinks

lukexor:


scrooloose:
Hi, the delay of almost a year! I'm having trouble seeing the effect of this patch. Can you tell me what problem this solves? Preferably with steps to reproduce the issue/error?

lukexor:
Wow, that has been awhile, I barely remembered submitting this.

It came down to the fact that expand("%:p") was not properly resolve for symlink paths when trying to utilize the findAndRevealPath function.

However, in trying to reproduce the original issue I had when I submitted this, I am not able to and expand seems to work just fine. You can go ahead and ignore this pull request.

PR #411, opened 2014-12-21
Fix for nonexistent dir/bookmarks given to NERDTree commands

xdl:
* Return early if exception was thrown in pathForString
* Stops a bunch of error messages getting printed onto the screen if you give the NERDTree, NERDTreeFromBookmark and NERDTreeToggle commands a dir/bookmark that doesn't exist (which also puts the plugin into an inconsistent state)


scrooloose:
Thanks :)

PR #417, opened 2015-02-13
Added NERDTreeNoSwitchTabs option.

mortonfox:
In NERDTree, if I select a file node and hit "o", it opens the file in a window in the current tab. However, if that file is already open in another tab, NERDTree switches to that tab to reuse the window. I find that this disrupts my workflow.

I asked if there's a way to prevent tab switching on [vi StackExchange](http://vi.stackexchange.com/questions/716/prevent-nerdtree-from-switching-tabs-when-opening-a-file-node), but there was no simple answer. So I decided to implement this option myself. And here are my changes.

### DIFF
--- a/doc/NERD_tree.txt
+++ b/doc/NERD_tree.txt
@@ -682,6 +682,9 @@ NERD tree. These options should be set in your vimrc.
                                 a buffer when a file is being deleted or renamed
                                 via a context menu command.
 
+|'NERDTreeNoSwitchTabs'|        Tells the NERD tree whether to reuse a window
+                                in another tab.
+
 ------------------------------------------------------------------------------
 3.2. Customisation details                             *NERDTreeOptionDetails*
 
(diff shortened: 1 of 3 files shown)


jamessan:
Shouldn't NERDTree just honor [`'switchbuf'`][1] here?

[1]: http://vimhelp.appspot.com/options.txt.html#%27switchbuf%27

mortonfox:
Yes, switchbuf does seem like what NERDTree should've followed. My guess is since it wasn't done that way from the start, changing the behavior to follow that option now would surprise a lot of folks.

scrooloose:
@mortonfox - I have added this commit: 87b27802b57bbf383a812b57338fbab28499b7e8

This essentially does the same as what you have done except that the option is given to the `Opener` object in its constructor instead of as a global var.

The way to use this is to override the `o` mapping. i.e. something like https://gist.github.com/scrooloose/0d8e626d3c7d538d695f

I am loath to add more global options to the nerdtree. It seems that everyone has their own preferences for details like this and I simply cant say yes to everyone. Ive concluded that offering an API for people to hack said details themselves is the best approach.

@jamessan - long time no see :) I had never read `:h switchbuf` till now. Respecting this option for the default `o` mapping could work. My concern is that changing the option would also change how quickfix windows work. I can see this being bad for some people. Thoughts?

jamessan:
@scrooloose _o/  I always forget that `'switchbuf'` only applies to specific commands instead of affecting the behavior of opening buffers in general.  In that light, I'm not sure there are applicable NERDTree actions that would map to what `'switchbuf'` affects.

I'm not sure I get your concern though.  Are you thinking people will change `'switchbuf'` to change NERDTree's behavior and then get confused when quickfix works differently?  If so, said people need to learn that it's much easier to just configure a map properly. ☺

amiorin:
I would prefer the option. I tried quickly the gist with latest version of NERDTree and it doesn't work. I didn't have time to investigate.

mortonfox:
@scrooloose The new Opener option does work for me. Thanks.

Since I use the mouse, I also needed to override the left mouse double-click. So in addition to what was in the key override gist, I added:

    call NERDTreeAddKeyMap({'key': '<2-LeftMouse>', 'callback': 'NERDTreeMyOpenFile', 'scope': 'FileNode', 'override': 1 })

lifecrisis:
This has been here too long. I vote to close.

PR #419, opened 2015-02-15
NERDTreeFind takes an optional filename argument

Arrowbox:
NERDTreeFind will default to the filename of the current buffer.
If the current buffer is a NERDTree buffer then the command will
reveal the file in the current tree instead of the primary one.

Allowing for an argument makes it easier for other plugins and user
functions to reveal a file in the NERDTree. I am using it with vim-vinegar
to reveal current file and not change the root directory of the NERDTree.

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -247,49 +247,63 @@ function! s:displayHelp()
 endfunction
 
 " FUNCTION: s:findAndRevealPath() {{{1
-function! s:findAndRevealPath()
-    try
-        let p = g:NERDTreePath.New(expand("%:p"))
-    catch /^NERDTree.InvalidArgumentsError/
-        call nerdtree#echo("no file for the current buffer")
-        return
-    endtry
+function! s:findAndRevealPath(file)
+
+    if empty(a:file)
+        try
+            let p = g:NERDTreePath.New(expand("%:p"))
+        catch /^NERDTree.InvalidArgumentsError/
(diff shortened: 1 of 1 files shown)


PhilRunninger:
@Arrowbox, I like this change. Can you resolve the conflicts in it, and update the doc file with the new functionality?

Thanks for the contribution.

PhilRunninger:
@Arrowbox, are you still interested in merging this PR? I think it would be a good addition. It just needs to have some conflict resolution and documentation updates.

@lifecrisis, what do you think of this one? There are a lot of whitespace changes in this because existing code was moved into an `else` block. I don't know if you've seen this trick or not.... To make comparison easier, at the expense of writing comments on lines of code, you can add `?w=1` to the **Files changed** URL, i.e.: https://github.com/scrooloose/nerdtree/pull/419/files?w=1

lifecrisis:
@PhilRunninger, that's a neat trick!

I'll have to look at this one later.  On the surface, things look good.

lifecrisis:
Looks like we can close this one. @PhilRunninger, when you review the other PR that does the same thing, you can close this at your discretion.

lifecrisis:
Hey @PhilRunninger, I noticed that this version of the feature uses `-complete=file` rather than `-complete=directory` like the other PR that does the same thing...

Do you think we should be using `-complete=file`?

PhilRunninger:
@lifecrisis, yes, that looks like the better way to go. I forgot to close this earlier, so I'll do that now.

PR #434, opened 2015-04-17
Use different arrow characters for compatibility.

handcraftedbits:
The "small arrow" characters used (u+25b8 and u+25be) don't seem to render properly on Windows using the Source Code Pro font in PuTTY (I've noticed this with other fonts as well).  On other systems they seem to render properly.  This patch uses larger arrow characters (u+25b6 and u+25bc) that seem to render consistently on Windows, Mac, and Linux.

### DIFF
--- a/autoload/nerdtree.vim
+++ b/autoload/nerdtree.vim
@@ -98,7 +98,7 @@ endfunction
 "FUNCTION: nerdtree#treeMarkupReg(dir) {{{2
 function! nerdtree#treeMarkupReg()
     if g:NERDTreeDirArrows
-        return '^\([▾▸] \| \+[▾▸] \| \+\)'
+        return '^\([▼▶] \| \+[▼▶] \| \+\)'
     endif
 
     return '^[ `|]*[\-+~]'
--- a/lib/nerdtree/tree_file_node.vim
+++ b/lib/nerdtree/tree_file_node.vim
@@ -383,13 +383,13 @@ function! s:TreeFileNode._renderToString(depth, drawText, vertMap, isLastChild)
         if self.path.isDirectory
             if self.isOpen
                 if g:NERDTreeDirArrows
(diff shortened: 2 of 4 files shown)


scrooloose:
Thanks. I have no easy way test this out on windows/osx but I'll take your word for it :)

vsushkov:
@handcraftedbits @scrooloose Unfortunately these symbols doesn't work quite well with OSX. See the screenshot.
screen shot 2015-05-02 at 12 50 16

KabbAmine:
Even in Linux with some fonts :worried:.
Here with a derivative of Ubuntu Mono:
nerdtree

scrooloose:
I have reverted these changes. Due to the rendering issues reported here and in comments here b3281cc63e44d97888a398d202f75c92ed0a0a04.

@handcraftedbits, not entirely sure what can be done to fix your situation. But, from the comments, it seems clear that the smaller arrows are less problematic overall.

kh3phr3n:
Thank you for changes, the problem with big arrows appears only in gVim, Urxvt looks good with the same font (Consolas), maybe a difference with letterspacing. but imo little arrows are just better :)

KabbAmine:
Even if the issue occurs only in gvim, the little arrows are better visually than big ones.
Thanks @scrooloose :+1:

vsushkov:
That's nice. Probably the best fix will be to introduce a config variable to set arrow character that user prefers. Like in airline:
```viml
let g:airline_left_sep = '»'
let g:airline_left_sep = '▶'
let g:airline_right_sep = '«'
let g:airline_right_sep = '◀'
```

handcraftedbits:
It's a shame it didn't work well everywhere.  Thanks for looking at it.  +1 to @vsushkov's suggestion to let us configure the arrows.

PR #447, opened 2015-05-09
custom filter mappings

albfan:
See https://github.com/Xuyuanp/nerdtree-git-plugin/pull/19 for implementation

For me it changes the shell workflow so now I prefer to navigate NERDTree than Ctrl-Z and git status. A total inflection point

f: Activate ignore filter (as usual)
fe (toggle exclude filter .git/info/exclude)
fi (toggle ignore filter .gitignore (not recursive by now)
fs (toggle status filter)

### DIFF
--- a/lib/nerdtree/creator.vim
+++ b/lib/nerdtree/creator.vim
@@ -3,6 +3,7 @@
 "buffer options and key mappings etc.
 "============================================================
 let s:Creator = {}
+let g:CustomIgnoreMappings = {}
 let g:NERDTreeCreator = s:Creator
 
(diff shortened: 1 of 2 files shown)


scrooloose:
Hmm, I have been thinking about this and concluded that it is too specialized and niche - I can't see anyone else using this functionality. I must be selective about what I pull so that the code base doesn't get too bloated. This was really the basis for creating the API in the first place.

Also, it is somewhat misleading as file filters are either all on, or all off. This pull request implies that they can be toggled individually - which may be true in your implementation, but is not the general case that the nerdtree code implies.

I will close this for now. If there is more demand then I will reopen.

PR #450, opened 2015-05-11
Optimized NERDTree when opening a large directory

vtsang:
- use sortKey to improve performance of sorting in s:TreeDirNode.sortChildren()
- use simple string comparison instead of regular expression in tree_dir_node.vim

### DIFF
--- a/autoload/nerdtree.vim
+++ b/autoload/nerdtree.vim
@@ -34,11 +34,11 @@ function! nerdtree#compareNodes(n1, n2)
     return a:n1.path.compareTo(a:n2.path)
 endfunction
 
-"FUNCTION: nerdtree#compareNodesBySortingToken(n1, n2) {{{2
-function! nerdtree#compareNodesBySortingToken(n1, n2)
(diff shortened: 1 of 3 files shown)


scrooloose:
I can confirm the speed results and I am keen to pull this.

However I have one question. You have

```
let s:format = "%0" . float2nr(ceil(log10(len(g:NERDTreeSortOrder)))) . "d"
...
let self.sortKey = printf(s:format, self.getSortOrderIndex()) . path
```

I have been trying to understand how this is different to simply having:

```
let self.sortKey = self.getSortOrderIndex() . path
```

Am I missing something?

vtsang:
This is to add leading zeros if number of elements in g:NERDTreeSortOrder > 10 so that you wont get 9 > 10 in string comparison.

scrooloose:
Ah yes, I see. I have merged this in and put this commit af85499edace2acca27ce41790ac333f62b689bb on top to rename a couple of variables - for clarity and convention.

Thanks :)

PR #453, opened 2015-05-22
Fixed incompatibility with older Vim versions

### LINKED ISSUE #452: `Unknown function: log 10`
I'm guessing it's since #450, but since recently I'm getting this error:

```
Error detected while processing /home/username/.vim/bundle/nerdtree/lib/nerdtree/path.vim:
line    5:
E117: Unknown function: log10
E116: Invalid arguments for function ceil(log10(len(g:NERDTreeSortOrder)))) . "d"
E116: Invalid arguments for function float2nr(ceil(log10(len(g:NERDTreeSortOrder)))) . "d"
E15: Invalid expression: "%0" . float2nr(ceil(log10(len(g:NERDTreeSortOrder)))) . "d"
```

WoLpH:
Fixed incompatibility with older Vim versions that are compiled without float function support such as `log10` and `ceil`

This fixes #452

### DIFF
--- a/lib/nerdtree/path.vim
+++ b/lib/nerdtree/path.vim
@@ -2,8 +2,11 @@
 "once here
 let s:NERDTreeSortStarIndex = index(g:NERDTreeSortOrder, '*')
 " used in formating sortKey, e.g. '%04d'
-let s:sortKeyFormat = "%0" . float2nr(ceil(log10(len(g:NERDTreeSortOrder)))) . "d"
-
+if exists("log10")
+    let s:sortKeyFormat = "%0" . float2nr(ceil(log10(len(g:NERDTreeSortOrder)))) . "d"
+else
+    let s:sortKeyFormat = "%04d"
+endif
 
 "CLASS: Path
 "============================================================


scrooloose:
I am currently away from the computer till Monday, so can't investigate this properly. However, if the log10 call is removed then this will cause the sort keys to be wrong under certain conditions.

@vtsang - can you comment further on this? Perhaps there is another way to generate the keys

WoLpH:
While that's true, the only case it could go wrong in some cases (not even all) is if you've got more than `10^4` (10000) files in a directory, seems like a sane enough default in case you have a Vim version without math methods :)

vtsang:
If Vim doesn't have float functions compiled with, then the sorting key may not work as expected. Having 4 digits as workaround is more than enough.

@WoLpH, the sortKeyFormat is not for number of files in a directory. It's for number of elements in g:NERDTreeSortOrder instead.
Its default value of NERDTreeSortOrder in my Vim is ['\/s', '*', '\.swpS', '\.bakS', '\~s'] which has only four elements. Not likely people defines it with more than 10k elements. So your workaruond should be pretty much fine in most of time.

PR #457, opened 2015-06-04
Option NERDTreeOpenTabAtTheEnd opens tabs at the end of the tab bar

schtibe:
Make it configurable to open a file in a tab after the last tab with the option g:NERDTreeOpenTabAtTheEnd.

What this basically does is calling `$tabnew` instead of calling `tabnew`.
See https://neovim.io/doc/user/tabpage.html#tab-page-commands for info


lifecrisis:
This is something that could easily be done with a custom mapping that wraps the default NERDTree functionality.  I move to close this pull request (which is quite dated, anyway).

@PhilRunninger, if you agree, feel free to close.

PR #477, opened 2015-08-19
Modified to set to calculate the initial value of g:NERDTreeWinSize.

magicdrive:
The initial value of g:NERDTreeWinSize is hard-coded, but not put the direct specific numerical value, the person who has been calculated by WindowSize I tried to fix it I think not to match in many cases

### DIFF
--- a/plugin/NERD_tree.vim
+++ b/plugin/NERD_tree.vim
@@ -88,7 +88,7 @@ if !exists('g:NERDTreeStatusline')
 
 endif
 call s:initVariable("g:NERDTreeWinPos", "left")
-call s:initVariable("g:NERDTreeWinSize", 31)
+call s:initVariable("g:NERDTreeWinSize", (&columns / 5))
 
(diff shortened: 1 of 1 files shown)


scrooloose:
This is an interesting idea but it has some flaws:

1. If you resize the vim window then the setting will be wrong
2. It fits the use case where you have the tree open all the time - but people who open the tree and then close it after switching files may not prefer this.

For now I will not pull this. I will reconsider if other people ask for this to be the default setting.

FYI, there is nothing stopping you from having this functionality by putting
```
let g:NERDTreeWinSize = (&columns / 5)
```
in your vimrc.

You might want to try something like this:

```
let g:NERDTreeWinSize = (&columns / 5)

autocmd vimresized * call s:updateNERDTreeSize()
function s:updateNERDTreeSize() abort
    let g:NERDTreeWinSize = (&columns / 5)
endfunction
```

PR #482, opened 2015-09-11
Added global variables for changing default arrows

PickRelated:
g:NERDTreeDirArrowExpandable
g:NERDTreeDirArrowCollapsible

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -105,3 +105,11 @@ Stick this in your vimrc:
 > Can I have different highlighting for different file extensions?
 
 See here: https://github.com/scrooloose/nerdtree/issues/433#issuecomment-92590696
+
+> How can I change default arrows?
+
+Use these variables in your vimrc. Note that below are default arrow symbols
+
+    let g:NERDTreeDirArrows = 1
+    let g:NERDTreeDirArrowExpandable = '▸'
+    let g:NERDTreeDirArrowCollapsable = '▾'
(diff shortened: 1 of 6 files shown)


PickRelated:
Please don't judge strictly. This is my first contribution. I am open to suggestions and critics.

scrooloose:
Thanks :) I imagine that certain symbols will mess the regexes up and will need to be escaped, but will deal with these if and when the bug reports come in.

PickRelated:
My pleasure. If someone complains - feel free to write to me. My email is in profile. I know how to deal with that. Though it will not bee too beautifull from code point of view.
For now I just tried to figure out what symbols are likely to show up and break the code. I avoided escaping where this could happen.

simonsmith:
I seem to be getting an error when trying to open NerdTree:

```
Error detected while processing /Users/simonsmith/.vim/plugged/nerdtree/syntax/nerdtree.vim:
line   26:
E121: Undefined variable: g:NERDTreeDirArrowCollapsible
E15: Invalid expression: 'syn match NERDTreeClosable #'.g:NERDTreeDirArrowCollapsible.'# containedin=NERDTreeDir,NERDTreeFile'
```

PickRelated:
I got a response from you on my email but I do not see the post. Is the problem still there? I can fix it.

simonsmith:
Apologies, I managed to resolve the issue after posting

PickRelated:
Okay, cool! I thought I have made a mistake of not defining variable but now I see that variable is defined on autoload. I guess relaunching vim after :PluginUpdate did the trick for you.

simonsmith:
Exactly yeah, I neglected to relaunch Vim :)

PR #496, opened 2015-10-04
Fix MacOS/neovim detection for extra menu options

### LINKED ISSUE #478: Neovim looses a lot of options
Not sure if this happen reported yet, but inside of neovim, a lot of the filesystem options are gone.

Here's a screen shot in mac-vim.

screen shot 2015-08-26 at 4 45 56 pm
 

Inside of neovim
screen shot 2015-08-26 at 4 46 42 pm

Worth noting that in terminal vim, I still have the same options as mac-vim, so this is neovim specific issue.

evindor:
Fixes #478, neovim is perfectly capable of executing those extra options.


scrooloose:
Merged, cheers :)

PR #502, opened 2015-11-02
Add ability to list file metadata...

JESii:
(Based on a change by github.com/apbarrero))
Add menu/list command to display file metadata
Support OSX as well as Unix


scrooloose:
Cheers, good idea :)

Issue #527, opened 2016-01-01
Feature Request: Keystroke to delete file's open buffer


PhilRunninger:
Many times, I'll be browsing through a codebase, not exactly sure of which file I need. I use the `go` keystroke to preview the file, and then move on to the next file to see if it's the file I'm looking for. Once I find the right file, I then have a bunch of buffers I no longer need.

A better workflow would be to have a keystroke I can type on a file node in the NERDTree to delete the buffer that contains the file. Then I could preview the file, and immediately close the buffer if it's the wrong one.

PhilRunninger:
This issue and its related pull request, #528, are redundant. Why did I do it that way? Anyway, the pull request is closed; no need to keep the issue open.

PR #528, opened 2016-01-01
Add bdelete functionality

PhilRunninger:
Implements #527 

Added a new key mapping `gd` to delete the current NERDTree file node's buffer, even if it's not currently active in a window. If that file is not open, there is no error. In all cases, an appropriate message is displayed, indicating success or failure.

Tested on console version only:
* Windows, vim 7.4
* Mac OSX, neovim 0.1.1


PhilRunninger:
I'm closing this request. I implemented this functionality as a NERDTree plugin: https://github.com/PhilRunninger/nerdtree-bwipeout-plugin

Issue #522, opened 2015-12-03
Missing dir arrows


dawikur:
I'm using Windows machine, ssh to Linux with putty and there running Vim.

`echo nerdtree#runningWindows()` returns `0`
but because I am rendering on Windows I don't have those fancy arrows. 
![arrows](https://cloud.githubusercontent.com/assets/11866902/11564746/10446b0c-99db-11e5-8f0e-0ff41537e2eb.png)

This was probably introduced with https://github.com/scrooloose/nerdtree/commit/677a83b2b63fda1c52205acd39973fe1b44e8b54.

Can we add option to force usage of +/~ ?

Workaround is simple:

``` vimL
let g:NERDTreeDirArrowExpandable="+"
let g:NERDTreeDirArrowCollapsible="~"
```

but this is always a workaround

ChaoyuChen:
Obviously that is a bug. Let's check NERD_tree.vim:

```
if !nerdtree#runningWindows()
    call s:initVariable("g:NERDTreeDirArrowExpandable", "▸")
    call s:initVariable("g:NERDTreeDirArrowCollapsible", "▾")
else
    call s:initVariable("g:NERDTreeDirArrowExpandable", "+")
    call s:initVariable("g:NERDTreeDirArrowCollapsible", "~")
endif
```
This section means 
When RunningWindows, arrows are +/~
Not RunningWindows, arrows are ▸/▾.

This logic is reversed.

So currently I remove the ! from the above code piece to make the logic right. 
```
if nerdtree#runningWindows()
    call s:initVariable("g:NERDTreeDirArrowExpandable", "▸")
    call s:initVariable("g:NERDTreeDirArrowCollapsible", "▾")
else
    call s:initVariable("g:NERDTreeDirArrowExpandable", "+")
    call s:initVariable("g:NERDTreeDirArrowCollapsible", "~")
endif
```
The change works for me.
![image](https://cloud.githubusercontent.com/assets/8033448/11697735/07529512-9ef6-11e5-97c1-4262d11c20ba.png)

ChaoyuChen:
Obviously this is a bug. Let's check plugin/NERD_tree.vim

```
if !nerdtree#runningWindows()
    call s:initVariable("g:NERDTreeDirArrowExpandable", "▸")
    call s:initVariable("g:NERDTreeDirArrowCollapsible", "▾")
else
    call s:initVariable("g:NERDTreeDirArrowExpandable", "+")
[...]

PR #539, opened 2016-01-26
Fixed windows arrow configuration if-statement

limwenbin:
Original if-statement was swapped for windows and non-windows, resulting in defaulting to fancy arrows for windows.


satori:
The original if-statement is correct, see #522

limwenbin:
oh right thanks for the help

PR #546, opened 2016-02-04
changed arrows for NERDTreeDirArrowExpandable and NERDTreeDirArrowCol…

gwilk:
…lapsible
Arrows used so far ('▸' unicode 9656 and '▾' unicode 9662) can't be find in popular monotype fonts like Courier New, Liberation Mono or Lucida Console (while available in DejaVu Sans Mono or Consolas)
Characters '►' (unicode 9658) and '▼' (unicode 9660) are much more popular.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -118,5 +118,5 @@ See here: https://github.com/scrooloose/nerdtree/issues/433#issuecomment-9259069
 
 Use these variables in your vimrc. Note that below are default arrow symbols
 
-    let g:NERDTreeDirArrowExpandable = '▸'
-    let g:NERDTreeDirArrowCollapsible = '▾'
+    let g:NERDTreeDirArrowExpandable = '►'
+    let g:NERDTreeDirArrowCollapsible = '▼'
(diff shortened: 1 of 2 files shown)


arturoherrero:
@gwilk @PhilRunninger This change has made my arrows very big in the terminal (Vim) or in MacVim. I really like to have the previous size that looks much better.

Before:
before

After:
after

arturoherrero:
Actually, I'm reading the README and it seems configurable, so probably I can configure in the `.vimrc` file. Anyway the default look and feel for a lot of users is going be as I've described.

scrooloose:
Hmm, we have gone back and forward over the arrows before and settled on smaller arrows.

I'm of the opinion that we should revert to small arrows (purely for aesthetic reasons) and add an FAQ with instructions to change to large arrows for fonts that dont support small.

Thoughts?

PhilRunninger:
@scrooloose , I agree. The smaller ones look better. When I'd merged this in, I misread the `if` statement and thought these arrows were a Windows-only problem. The FAQ is already in the README. I'll go ahead and revert the merge if you give the thumbs up.

PhilRunninger:
@gwilk, I've reverted this pull request. To use the large arrows, you'll have to override them as shown in the FAQ section of the README.

scrooloose:
@PhilRunninger all good mate, thanks for fixing.

Oh, I didn't spot that FAQ... I really need to learn to read :smile:

gwilk:
OK. My point was, that the small arrows are missing in many popular fonts. Perhaps the fonts will be updated some day. For now, changing configuration is good enough.

Issue #537, opened 2016-01-24
Dir arrows not functioning properly


fungl164:
Using any of these in .vimrc
```code
let g:NERDTreeDirArrowExpandable = ' '
let g:NERDTreeDirArrowCollapsible = ' '
```
Or
```code
let g:NERDTreeDirArrowExpandable = ''
let g:NERDTreeDirArrowCollapsible = ''
```
causes NERDTree not to expand after two levels.
<br>
On the other hand, using
```code
let g:NERDTreeDirArrowExpandable = '+'
let g:NERDTreeDirArrowCollapsible = '-'
```
causes 
```code
E16: Invalid range
E16: Invalid range
E475: Invalid argument: NERDTreeFile  #^[^"\.-+] *[^-+]*# contains=NERDTreeLink,NERDTreeRO,
NERDTreeBookmark,NERDTreeExecFile
```

adityanatraj:
I had the same issue. As a temporary solution, I've been using `"X"` for Exapandable and `"O"` for Collapsible. 

Looking at the way the regular expression (`#^[^"\.-+] *[^-+]*#`) is being generated [here](https://github.com/scrooloose/nerdtree/blob/master/syntax/nerdtree.vim#L31), it would break the generation to use the `-` character. 

Within `[ ]`'s, the `-` is a protected character meaning "a range of characters" (like `0-9`, `a-z`, etc). Since, the expansion doesn't worry about _escaping_ (that is to inform the code that this letter is used as-is by prepending a `\`), using the `-` as either the expanding or collapsing character would be no good. In fact, any regex special character or term would be a bad choice.

I don't really know any vimscript, so I can't fix it, but that's the (or an) issue methinks.

mdu4rt3:
Please bring back the option

let g:NERDTreeDirArrows=0

This used to work fine, now it's gone, not sure why it was removed.

xiongm:
on the latest version, same error. annoying

Error detected while processing dotfiles/vim/plugged/nerdtree/syntax/nerdtree.vim:
line   31:
E16: Invalid range
E475: Invalid argument: NERDTreeFile  #^[^"\.-+] *[^-+]*# contains=NERDTreeLink,NERDTreeRO,NERDTreeBookmark,NERDTr
eeExecFile

DeepikaNR:
Namaste Folks, 

Firstly, Congo on awesome work!!
I faced similar issues as elicited above but upon mapping to '>' and '<' things seem to be working fine.

[...]

PR #548, opened 2016-02-06
Issue #537: Fix dir arrows not functioning properly

adityanatraj:
added characters that are special when within character
classes in a regular expression to the `escape(...)` call
when building `s:dirArrows`.

this does not fix when `' '` or `''` are your `dirArrows` and you
can't open up subdirs. i think that's an issue with how nerdtree
distinguishes things to traverse in the filetree.

this fixes: https://github.com/scrooloose/nerdtree/issues/537

### DIFF
--- a/syntax/nerdtree.vim
+++ b/syntax/nerdtree.vim
@@ -25,7 +25,7 @@ syn match NERDTreeDirSlash #/# containedin=NERDTreeDir
 exec 'syn match NERDTreeClosable #'.escape(g:NERDTreeDirArrowCollapsible, '~').'# containedin=NERDTreeDir,NERDTreeFile'
 exec 'syn match NERDTreeOpenable #'.escape(g:NERDTreeDirArrowExpandable, '~').'# containedin=NERDTreeDir,NERDTreeFile'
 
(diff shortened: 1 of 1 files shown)


mdu4rt3:
Please bring back the option

let g:NERDTreeDirArrows=0

This used to work fine, now it's gone, not sure why it was removed.

adityanatraj:
what did
```
let g:NERDTreeDirArrows=0
```
do ?

I'm sorry, I don't know what that did, I'm relatively new to NERDTree

mdu4rt3:
It disabled the ascii arrows in favor of + and - for directories.

adityanatraj:
@mdu4rt3 I cannot add that option as feature changes/behaviors are outside the scope of this PR. This is simply to fix the current issue with the broken characters for dir arrows. 

If you notice within the [CHANGELOG here](https://github.com/scrooloose/nerdtree/blob/master/CHANGELOG#L5), the behavior you requested was purposefully removed.

PhilRunninger:
@adityanatraj , Thanks for providing this fix.

PR #550, opened 2016-02-09
Make window creation command configurable

wincent:
This commit adds a `NERDTreeCreatePrefix` setting that can be used to
prefix the `:edit` command that is used to create the NERDTree tree
window. Defaults to "silent", meaning that out of the box the window
will be created with "silent edit".

Users may wish to configure this to produce other effects. For example,
`NERDTreeCreatePrefix` can be set to "silent keepalt keepjumps" in order
to make NERDTree create its window with "silent keepalt keepjumps edit".

This can be used to create an effect analogous to the `g:netrw_altfile`
setting in netrw. An example of why you might want to do this is
described here:

    https://github.com/tpope/vim-vinegar/issues/25

I'm not using vim-vinegar myself, but I am using something like it here:

    https://github.com/wincent/wincent/blob/3efaa8fa50895/roles/dotfiles/files/.vim/plugin/mappings.vim#L60

And having `NERDTreeCreatePrefix` enables me to map "-" to show the
current file in context, and `^-6` to jump back to it.

### DIFF
--- a/doc/NERD_tree.txt
+++ b/doc/NERD_tree.txt
@@ -679,6 +679,9 @@ NERD tree. These options should be set in your vimrc.
                                 a buffer when a file is being deleted or renamed
                                 via a context menu command.
 
+|'NERDTreeCreatePrefix'|        Specify a prefix to be used when creating the
+                                NERDTree window.
+
 ------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


scrooloose:
Sounds like a good use case to me. Cheers :)

PR #556, opened 2016-02-27
Allow user-defined read-only indicator.

acarlson1029:
(inspired by overwriting g:NERDTreeDirArrowCollapsible/Expandable)

Replace instances of the RO string with a variable that the user can
override. Useful for custom unicode glyphs, i.e. Font Awesome.

* Initialize variable g:NERDTreeGlyphReadOnly = "RO".
   -> plugin/NERD_tree.vim
* Replace instances of 'RO' with g:NERDTreeGlyphReadOnly

Preview:
screen shot 2016-02-26 at 6 02 12 pm

> **NOTE** - I left the square brackets around the indicator so users don't have to deal with issues escaping characters.

### DIFF
--- a/lib/nerdtree/path.vim
+++ b/lib/nerdtree/path.vim
@@ -61,7 +61,7 @@ function! s:Path.cacheDisplayString() abort
     endif
 
     if self.isReadOnly
-        let self.cachedDisplayString .=  ' [RO]'
+        let self.cachedDisplayString .=  ' ['.g:NERDTreeGlyphReadOnly.']'
     endif
 endfunction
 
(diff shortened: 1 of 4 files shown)


scrooloose:
Cheers. That padlock glyph looks awesome!

acarlson1029:
Thanks @scrooloose !

PhilRunninger:
@acarlson1029 , how do you specify the glyphs in your vimrc? I found the Unicode values for each:

```
1F4C1 - closed file folder
1F4C2 - open file folder
1F512 - lock
```

But vim wants exactly 4 digits when I try to insert the character with the following keystrokes:

```
<Esc>$i<C-V>u1f512<Esc>
```

I just end up with a garbage character and a 2. I see the same behavior when encoding is set to the default (utf-8), as well as when setting it to utf-16.

I looked at Font Awesome, but I wasn't quite sure how to use it and one of the Powerline fonts for Airline at the same time.

acarlson1029:
@PhilRunninger, if I remember correctly, I just copy-pasted the glyphs into my vim buffer from somewhere else (like the FontAwesome webpage).
I ran into the same issues as you when trying to type the glyphs manually.

PR #558, opened 2016-03-05
Add option to disable collapsing of directory names

juanibiapina:
This pull request adds the following option: `NERDTreeCascadeSingleChildDir`,
with default value of 1 (in order to keep the current behavior).

It can be disabled to prevent directory names from being collapsed into the same
line.

It doesn't affect the behavior of `NERDTreeCascadeOpenSingleChildDir`, because
it only changes how directories are displayed, not opened. Both options can be
used together in any combination.

After making the changes I realized "cascading" can generally be replaced with
"collapse", except when referring to "cascade open". I decided not to make this
change on this pull request.

### DIFF
--- a/doc/NERD_tree.txt
+++ b/doc/NERD_tree.txt
@@ -668,14 +668,18 @@ NERD tree. These options should be set in your vimrc.
 |'NERDTreeWinSize'|             Sets the window size when the NERD tree is
                                 opened.
 
-|'NERDTreeMinimalUI'|           Disables display of the 'Bookmarks' label and 
+|'NERDTreeMinimalUI'|           Disables display of the 'Bookmarks' label and
                                 'Press ? for help' text.
+
+|'NERDTreeCascadeSingleChildDir'|
+                                Collapses on the same line directories that
+                                have only one child directory.
 
 |'NERDTreeCascadeOpenSingleChildDir'|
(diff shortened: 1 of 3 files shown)


juanibiapina:
@scrooloose

timofonic:
@frpz fork is alive, at least...
https://github.com/frpz/nerdtree

ksmithbaylor:
I'd love to see this option! I've been going crazy trying to bring back this behavior.

shannonmoeller:
@scrooloose Please merge this, or implement similar.

https://github.com/scrooloose/nerdtree/issues/76

nmhnmh:
@scrooloose please merge this, I really need this

hookdump:
Thank god this option exists. How is "1" the default value for this? The option does not work. I cannot open directories that have been collapsed.

lifecrisis:
@hookdump, you don't need to open directories that have been collapsed *because there is nothing else in them.*  This is a very useful feature in a number of programming contexts.  It still has bugs, but collapsing directories by default is *not* one of them.

juanibiapina:
@hookdump The default value is 1 to keep backwards compatibility.

shannonmoeller:
@PhilRunninger Thank you for merging this! Hadn't noticed until @hookdump commented. Soooo much better.

PR #574, opened 2016-04-25
Remove empty segments

sooth-sayer:
In some cases 'fullpath' may be something like 'some_path/some_path**//**some_path'.


scrooloose:
This is strange indeed :-/ Can you give an example that will reproduce this?

It sounds more like a symptom of a bug in either of these above lines:

```
    call self.extractDriveLetter(a:fullpath)
    let fullpath = s:Path.WinToUnixPath(a:fullpath
```

sooth-sayer:
No, vim expand(...) may return it.
In my case the path like 'some_path/some_path//some_path' is in the tag file.
Vim handles it as valid file path.

PR #582, opened 2016-05-13
Expose initNerdTreeInPlace

dmerejkowsky:
In need that to make NerdTree work with neovim's new TabNewEntered event.

My goal is to set the working directory any time a new tab is opened.

Unfortunately with NERDTreeHijackNetrw=1 things do not work as expected, so I had to set it to false, and then manually create a NERDTree window, like this:

``` vim

let g:NERDTreeHijackNetrw=0

function! OnTabEnter(path)
  if isdirectory(a:path)
    let dirname = a:path
    execute ":NERDTreeInPlace " . dirname
  else
    let dirname = fnamemodify(a:path, ":h")
  endif
  execute "tcd ". dirname
endfunction()

autocmd TabNewEntered * call OnTabEnter(expand("<amatch>"))
```

Note that I can't use plain old `:NERDTree` because it tould create a _new_ window

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -548,6 +548,7 @@ endfunction
 " FUNCTION: nerdtree#ui_glue#setupCommands() {{{1
 function! nerdtree#ui_glue#setupCommands()
     command! -n=? -complete=dir -bar NERDTree :call g:NERDTreeCreator.CreateTabTree('<args>')
+    command! -n=1 -complete=dir -bar NERDTreeInPlace :call s:initNerdTreeInPlace('<args>')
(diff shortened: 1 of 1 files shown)


dmerejkowsky:
Few notes:

* There may be a better way to call `s:initNerdTreeInPlace` from outside the plugin ...
* If you like this new command, I'll be happy to update the documentation too

Thanks :)

PhilRunninger:
@dmerejkowsky , your changes refer to an old version of NERDTree. `initNerdTreeInPlace` hasn't been part of the code since January 2013. Pull down the latest version of NERDTree and see if you still need this change. If so, you can submit a new pull request. This one I will close.

dmerejkowsky:
I no longer need it :)

Issue #519, opened 2015-12-03
Can't reopen after use 'x' while the directory only have one child directory.


xinyifly:
/a
/a/b
/a/b/c
/a/b/d
So, I open a. Get 'a/b', list 'c,d'.
Then, I close using x. Get 'a'
After, can't reopen 'a'.

czak:
I'm having the same issue. After closing a "cascade" dir via <kbd>x</kbd>, it won't show the contents anymore.

Here's an asciicast displaying the issue:

<a href="https://asciinema.org/a/ef4wk8hz5mk9hub6msjlaflvb" target="_blank"><img src="https://asciinema.org/a/ef4wk8hz5mk9hub6msjlaflvb.png" /></a>

The issue can be somewhat resolved by setting:

```
let NERDTreeCascadeOpenSingleChildDir=0
```

...however this disables the automatic cascade open (as per the option name), which is convenient for e.g. Java projects.

czak:
I've been trying to track down this issue.
I don't have a fix yet, but before I forget here are my notes.

(please excuse imprecise terms, Vimscript is new to me)

The difference I'm seeing **before** and **after** closing a cascade node with `x` manifests here:

https://github.com/scrooloose/nerdtree/blob/master/lib/nerdtree/ui.vim#L443

The output of `self.nerdtree.root.renderToString()` **before** `x`-ing the `c` node is:

```
▾ a/b/
    c
```

The output of `self.nerdtree.root.renderToString()` **after** `x`-ing the `c` node is only:

```
▾ a/b/
```

(the nodes I'm referring to are the `a/b/c` folder structure as linked in my asciicast).

I believe some of the flags set on the nodes during [closeCurrentDir](https://github.com/scrooloose/nerdtree/blob/master/autoload/nerdtree/ui_glue.vim#L186-L205) prevent the last child from being rendered.

I'll do some more digging, but maybe this helps someone.

czak:
If someone could help me how to set a breakpoint on this particular function, that would be great:

https://github.com/scrooloose/nerdtree/blob/master/lib/nerdtree/tree_file_node.vim#L325

My current solution is:

* I turned the script-local `s:activateDirNode` into a global `nerdtree#ui_glue#activateDirNode`
* I set `:breakadd func nerdtree#ui_glue#activateDirNode`
* I step through until I reach that line
This is a PITA

`TreeFileNode.renderToString` is different because it's in the dictionary an I can't seem to get a breakpoint in there.

[...]

PR #591, opened 2016-06-22
Fix upward traversal when closing a cascade

czak:
This is my fix for #519.

I noticed that `closeCurrentDir` was more "destructive" than a regular node toggle.
This was due to the "upward traversal" loop closing every parent and parent's parent and so on.

My change ensures that only a single node in a cascade is closed.
The appropriate node is found by traversing upwards.

The traversal stops when either happens:

* a node is found with more than 1 child
* a node if found with a child that is not a directory

That node becomes the `parent`, the loop `break`s and the regular closing process continues.

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -187,17 +187,17 @@ endfunction
 " closes the parent dir of the current node
 function! s:closeCurrentDir(node)
     let parent = a:node.parent
+    while g:NERDTreeCascadeOpenSingleChildDir && !parent.isRoot()
+        let childNodes = parent.getVisibleChildren()
+        if len(childNodes) == 1 && childNodes[0].path.isDirectory
+            let parent = parent.parent
+        else
+            break
+        endif
+    endwhile
     if parent ==# {} || parent.isRoot()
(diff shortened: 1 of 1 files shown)


czak:
Okay I noticed one more issue - a `closeCurrentDir` call on a root-level cascade closes the incorrect directory. Looking for a fix now.

czak:
Okay I resolved the issue in 2924ab2. I'm now satisifed with the fix.

The change now makes `closeCurrentDir` work as follows:

1. Traverse upwards to a nearest "closeable" `parent`
  * this will be a no-op when we're not in a cascade, because it will `break` on the first iteration
2. Now check if `parent.isRoot()`
  * previously this was the first step, but now we must confirm after the traversal
3. Continue with the closing

Let me know what you think!

PhilRunninger:
Looks good, @czak . 👍

mckayward:
Thanks @czak! This has been bothering me a ton lately. I'll be using your branch until this gets merged.
:+1:

scrooloose:
Thanks @czak, I've been noticing this myself lately but have been too busy to fix :)

Thanks to others for testing :)

czak:
Thanks for a quick response everyone :) 👍

PR #595, opened 2016-06-30
Use fnameescape to quote filenames for :badd.

zacharyvoase:
When you move a file in NERDTree's FS menu, and a buffer is already open with it, you're prompted to delete the old buffer and replace it with a new buffer pointing to the file's new location. This wasn't working correctly for me: because the new filename is passed to `badd` quoted with single quotes, Vim just created a new buffer pointing to a file called `'/Users/zack/.../filename'` _with_ the quotes. See the odd results of this when trying to rename a file in the nerdtree source tree:

screenshot 2016-06-30 16 20 00

This patch uses the `fnameescape()` function to do the filename escaping, rather than just surrounding with single quotes. I've tested it and it fixes the error seen in the above screenshot.


PhilRunninger:
👍  I've noticed this before. Thanks for the fix, @zacharyvoase.

Issue #62, opened 2011-04-25
Hide executables on linux


drewm1980:
This is a feature request:

Add some facility for hiding executables from the tree.  On windows one can use a file filter on .exe, but on Linux, programs generally have no extension.  Executable shell scripts should remain visible.

Thanks!

schickling:
This seems to be no longer supported. What's the current way to do this @scrooloose ?

PhilRunninger:
@schickling , see `:h NERDTreeAddPathFilter`. This function lets you specify your own callback function for complex filtering needs.

PR #604, opened 2016-07-25
New option: NERDTreeShowExecutableFlag

### LINKED ISSUE #590: Hide * in executable files
Is there any configuration to hide the * in executable files on NERDTree? Do they come from NETrw? I saw the source code of NETrw and doesn't look like there is any configuration available to hide them. They cause bugs in some nerdtree plugins.

PhilRunninger:
This pull requests fixes #590. A new option `NERDTreeShowExecutableFlag` can be set to 0 in your .vimrc to prevent displaying the trailing asterisk of executable files.

### DIFF
--- a/doc/NERD_tree.txt
+++ b/doc/NERD_tree.txt
@@ -593,7 +593,9 @@ file explorers have.
 The script comes with two default menu plugins: exec_menuitem.vim and
 fs_menu.vim. fs_menu.vim adds some basic filesystem operations to the menu for
 creating/deleting/moving/copying files and dirs. exec_menuitem.vim provides a
-menu item to execute executable files.
+menu item to execute executable files. (By default, executable files are
+displayed with a trailing asterisk. This can be turned off by setting the
+|'NERDTreeShowExecutableFlag'| option to 0.)
 
 Related tags: |NERDTree-m| |NERDTreeApi|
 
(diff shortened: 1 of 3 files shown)


tiagofumo:
thanks @PhilRunninger, I home it gets merged :smile:

PhilRunninger:
The problem this PR addresses has been resolved a different way, so I'm going to close this without merging.

manifoldQAQ:
Hi @PhilRunninger , you mentioned a workaround for this issue, could you elaborate it a little bit? I have a lot of executables after a makefile project is built and they are messing up my NERDTree list. Thanks!

PhilRunninger:
From what I can piece together from the comments above and in #590, @tiagofumo's plugin, [vim-nerdtree-syntax-highlight](https://github.com/tiagofumo/vim-nerdtree-syntax-highlight), had a hard time with the asterisks, but only because those files were colored with a different syntax than non-executable files. He made some changes to his plugin, and no longer needed the change to be made in NERDTree.

manifoldQAQ:
Oh, I've mistaken the meaning of NERDTreeShowExecutableFlag. What if I don't want to display all executables under a directory? I searched through the issue list and there's one issue related to this #62, which seems to no longer work. Is there any solution? Thanks! @PhilRunninger

PhilRunninger:
@manifoldQAQ, see `:h NERDTreeAddPathFilter`. This function lets you specify your own callback function for complex filtering needs.

PhilRunninger:
There is a problem with the code as shown in the help file. vim complains that `NERDTreeAddPathFilter` is undefined. This has to do with the order in which plugins (including `vimrc`) are executed. To solve that, I did this instead.
```
augroup NTStartup
    autocmd!
    autocmd VimEnter * call NERDTreeAddPathFilter('MyFilter')
augroup END

function! MyFilter(params)
    return a:params['path'].isExecutable
endfunction
```
This causes `call NERDTreeAddPathFilter` to be called after the function has been created by the NERDTree plugins.

This function will hide all executables in all directories. If for some reason, you want to see them, you can use the `f` keystroke (or whatever you assign to `NERDTreeMapToggleFilters`) to toggle them.

PR #608, opened 2016-08-04
added NERDTreeFocusClose command

chdupre:
I've added a command called NERDTreeFocusClose. 
Before => After
Tree is closed => Tree opens
Tree is opened and focused => Tree closes
Tree is opened but not focused => Tree is focused


PhilRunninger:
@chdupre Rather than add another command to NERDTree, this functionality can be done with a one-liner in the vimrc:
```
nnoremap <silent><expr> <leader>t (winnr()==g:NERDTree.GetWinNum() ? ":NERDTreeClose\<CR>" : ":NERDTreeFocus\<CR>")
```

PR #623, opened 2016-10-03
Fix typo in argument usage in promptToRenameBuffer

dschaub:
Was seeing an issue where renaming a file and then reloading it would throw a pretty gnarly error. Tracked it down to this typo, which seems to fix it.

Related issue: https://github.com/scrooloose/nerdtree/issues/622


scrooloose:
Excellent, thanks :)

Issue #602, opened 2016-07-22
can't check arrow char in mingw32


zzs213:
because mingw32 has no full support for unicode, and nerdtree just check win16, wwin32 and win64. So when try to set "▸" , E114 appear.
Maybe need to check has('win32unix') too. I'am not sure. Because of the function runningWindows() was called many place.

PhilRunninger:
@zzs213, do you feel comfortable running git commands in your nerdtree folder? If so, you can do some testing for me, since I don't have access to cygwin. Go to your nerdtree folder, and type these commands:

```
git pull
git checkout win32unix
```

Then go into vim, and see if the error has been fixed.

Thanks!

PR #626, opened 2016-10-06
Determine if we're running in Cygwin on Windows.

PhilRunninger:
Solves #602.

### DIFF
--- a/autoload/nerdtree.vim
+++ b/autoload/nerdtree.vim
@@ -103,7 +103,7 @@ endfunction
 
 "FUNCTION: nerdtree#runningWindows(dir) {{{2
 function! nerdtree#runningWindows()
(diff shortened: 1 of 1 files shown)


sst6z:
This completely breaks NERDTree on Cygwin.
(When opening nerdtree it just says "NERDTree: No bookmark or directory found for:".)
My guess is that now NERDTree uses Windows style paths etc. in Cygwin as well, but Cygwin normally needs Unix style paths.
(BTW: On my Cygwin installation with the correct font the small arrow symbols work just fine.)

Therefore I'd suggest reverting this commit until a proper fix is found.
I'd be glad to test it on my machine (running vim in mintty/cygwin).

PhilRunninger:
@sst6z, thanks for the feedback. The change has been reverted.

PR #629, opened 2016-10-11
Add option to set bookmarks file field separator.

Doron-Behar:
Useful for ranger (console file-manager) users.


Doron-Behar:
It doesn't work yet, I'll have to work on it a bit more and than I'll open a new pull-request.

Issue #409, opened 2014-12-18
Arrow color changed


aaronmoodie:
![screen shot 2014-12-17 at 7 31 35 pm](https://cloud.githubusercontent.com/assets/105030/5490099/f263fc94-869c-11e4-9399-8b9021268e1a.png)

I've just done a clean setup of Vim, and after installing NERDTree, noticed that the arrows on the directories are now a separate color, regardless of theme.

Leeiio:
+1

PhilRunninger:
Closes I don't see anything in the history to indicate the syntax had changed. Regardless, I like the change submitted by @Leeiio .
![image](https://cloud.githubusercontent.com/assets/5598066/19363988/4fc9a2b6-915a-11e6-9fc0-5d54a5462f74.png)

PR #630, opened 2016-10-13
Update NERDTreeOpenable and NERDTreeClosable color

Leeiio:
It is more comfortable

### DIFF
--- a/syntax/nerdtree.vim
+++ b/syntax/nerdtree.vim
@@ -72,8 +72,8 @@ hi def link NERDTreeDir Directory
 hi def link NERDTreeUp Directory
 hi def link NERDTreeFile Normal
 hi def link NERDTreeCWD Statement
-hi def link NERDTreeOpenable Title
-hi def link NERDTreeClosable Title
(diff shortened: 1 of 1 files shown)


PhilRunninger:
I don't see anything in the history to indicate the syntax had changed. Regardless, I like the change submitted by @Leeiio .
image

svanharmelen:
@PhilRunninger can we revert this? I really don't agree that this should be more comfortable. Having it be exactly the same color as the dir names make it less clear to my eyes...

Leeiio:
@svanharmelen Maybe you can change the arrow color like this put in your .vimrc right now:
hi NERDTreeOpenable cterm=NONE ctermbg=yellow ctermfg=black guifg=#000000 guibg=NONE
hi NERDTreeClosable cterm=NONE ctermbg=yellow ctermfg=black guifg=#000000 guibg=NONE

I change the arrow color because the #409 metioned.

svanharmelen:
Check... Thanks for the info.

Issue #420, opened 2015-02-18
Cannot Open Files Whose Filename Includes a Dollar Sign


dwtj:
If I try to use `nerdtree` to open a file whose file name includes a dollar sign, that file is not necessarily opened correctly. Consider the following example:

![screen shot 2015-02-17 at 11 43 25 pm 2](https://cloud.githubusercontent.com/assets/2457848/6243007/e029e196-b6fe-11e4-950c-bddd9b5241eb.png)

When the first `.java` file in the `nerdtree` view is opened (either in the previous window, a vertical split, or a horizontal split), that file is not opened in the expected window. Instead, that window opens the wrong file (potentially one which doesn't exist). As you can see in the status line in the example, `type/Panini.java` was opened instead of `type/Panini$Duck$Array$Types.java`.

In case it is helpful, I am running:
- Mac OS X 10.10.2
- `Vim` 7.4.488
- `tmux` 1.9a
- `scrooloose/nerdtree` commit 3b98a7fcae8f9fff356907171f0406ff8cd28921 (installed via `Vundle`)
- `jistr/vim-nerdtree-tabs` commit 0decec122e9bb3e9328b01fa20a9650e79cc6ca7 (installed via `Vundle`)
- Other `Vundle`-installed `Vim` plugins, though they probably aren't relevant.

dwtj:
Note that in the given example, the same incorrect behavior happens with the `Vim` command

```
:e types/Panini$Duck$Array$Types.java
```

This will (maybe unexpectedly) start editing `types/Panini.java`.

One can escaping each dollar sign to make the `:e` command open the expected file.

```
:e types/Panini\$Duck\$Array\$Types.java
```

yan12125:
Just a bump. Any progress on this issue? I'm encountering the same problem.

yan12125:
Here's a quick hack:
```Diff
diff --git a/lib/nerdtree/path.vim b/lib/nerdtree/path.vim
index c0242d0..88149de 100644
--- a/lib/nerdtree/path.vim
+++ b/lib/nerdtree/path.vim
@@ -266,7 +266,9 @@ endfunction
 
 "FUNCTION: Path.edit() {{{1
 function! s:Path.edit()
-    exec "edit " . self.str({'format': 'Edit'})
+    let edit_command = "edit " . self.str({'format': 'Edit'})
+    let edit_command = substitute(edit_command, '\$', '\\$', '')
+    exec edit_command
 endfunction
 
[...]

PR #649, opened 2016-12-29
fix bug with files and directories that have dollar sign in name

alegen:
this is a fix for issue #420 which seems to be the better than the fix-hack proposed in that thread.


PhilRunninger:
@alegen Good work. Thanks for the contribution.

PR #655, opened 2017-01-26
Update fs_menu.vim

olivier-o:
add `h` option to the `ls` command, for better readability.

### DIFF
--- a/nerdtree_plugin/fs_menu.vim
+++ b/nerdtree_plugin/fs_menu.vim
@@ -203,7 +203,7 @@ endfunction
 function! NERDTreeListNode()
     let treenode = g:NERDTreeFileNode.GetSelected()
     if treenode != {}
-        let metadata = split(system('ls -ld ' . shellescape(treenode.path.str())), '\n')
+        let metadata = split(system('ls -ldh ' . shellescape(treenode.path.str())), '\n')
         call nerdtree#echo(metadata[0])
     else
         call nerdtree#echo("No information avaialable")


nerdtree_plugin/fs_menu.vim
```diff
@@ -203,7 +203,7 @@ endfunction
 function! NERDTreeListNode()
     let treenode = g:NERDTreeFileNode.GetSelected()
     if treenode != {}
-        let metadata = split(system('ls -ld ' . shellescape(treenode.path.str())), '\n')
+        let metadata = split(system('ls -ldh ' . shellescape(treenode.path.str())), '\n')
```
PhilRunninger: @olivier-o I actually like seeing the exact size of the file, but to make it easier to read, what do you think of this option?

```
let metadata = split(system("ls -ld " . shellescape(treenode.path.str()) . " | sed -e :a -e 's/\(.*[0-9]\)\([0-9]\{3\}\)/\1,\2/;ta'"), '\n')
```

Props to [Brian Cooper](http://stackoverflow.com/users/695419/brian-cooper) on StackOverflow for this [answer](http://stackoverflow.com/a/23666338/510067).

olivier-o: @PhilRunninger, the sed option doesn't work very well with the proposed regex

```
ls -lh
805K 17 Feb 10:14 2017-02-14-Windsor-T21986.jpeg
```
with sed and the proposed regex the file name get corrupted
```
ls -l | sed -e :a -e 's/\(.*[0-9]\)\([0-9]\{3\}\)/\1,\2/;ta'
823,989 17 Feb 10:14 2,017-02-14-Windsor-T21,986.jpeg
```

PhilRunninger: Yeah, I suspected that would happen. I was trying to develop a better regex. It works with grep, but not sed. Really strange. Assuming it can be fixed to work with sed, it will also comma-fy the years, so this ultimately isn't a solution either.

```
ls -l | grep '\(^\|\s\)\(\d\+\)\(\d\{3\}\)\(\(,\d\{3\}\)*\)\(\s\|$\)'
ls -l | sed -e :a -e 's/\(^\|\s\)\(\d\+\)\(\d\{3\}\)\(\(,\d\{3\}\)*\)\(\s\|$\)/\1\2,\3\4\6/;ta'
```

olivier-o: would be possible to use stat and ls -lh ? 
```
stat -f "%z" "2017-02-14-Windsor-T21986.jpeg"
823989
``` 
returns only the file size in bytes. but not sure if the formatting options are consistent accros OSX and Linux

PhilRunninger: I'm back. Sorry for the delay. What if we combine these ideas? It's ugly, but it could work. I don't have access to Linux, just OSX.

### In a bash shell...
```
x=$(stat -f "%z" vimrc) && y=$(echo $x | sed -e :a -e 's/\(.*[0-9]\)\([0-9]\{3\}\)/\1,\2/;ta') && ls -ld vimrc | sed -e "s/ $x / $y /" && unset x && unset y
```

### In vimscript...

```
let metadata =
\   split(
\     system('x=$(stat -f "%z" ' .
\            shellescape(treenode.path.str()) .
\            ') && y=$(echo $x | sed -e :a -e "s/\(.*[0-9]\)\([0-9]\{3\}\)/\1,\2/;ta") && ls -ld ' .
\            shellescape(treenode.path.str()) .
\            ' | sed -e "s/ $x / $y /" && unset x && unset y'),
\       '\n')
```

PhilRunninger:
@olivier-o, any updates on this? What do you think of my last proposal?

PhilRunninger:
@olivier-o , I'm going to close this pull request, effectively replacing it with my own #754.

Issue #663, opened 2017-02-25
Is there any shortcut to toggle between my editor screen and nerd-tree tab in vim?


balogic:

lifecrisis:
As you noted there are plenty of ways to map this kind of behaviour yourself. Mapping `<C-N>` is already mentioned in the `README.markdown` file, and the newly documented `:NERDTreeFocus` command can be used for this purpose as well.

These types of maps are really up to the individual user, so you'll have to experiment with the macros that you might want. Generally, that will be some combination of mappings involving `:NERDTreeToggle` and `:NERDTreeFocus`.

I'm recommending to @PhilRunninger that this issue be closed since no real issue with the NERDTree is reported here.

PR #664, opened 2017-02-25
Added help info and shortcut in readme

balogic:
The help info and shortcut for switching tabs are added in
README.markdown.


balogic:
Solving the issue #663

lifecrisis:
@PhilRunninger, I recommend that this PR be rejected. The suggestion for `<C-n>` is already available and complete in `README.markdown`. In addition, telling readers here about the `?` mapping does nothing since the suggestion is already visible within the NERDTree window itself.

PR #671, opened 2017-03-10
Update NERD_tree.vim

FinallyFinancialFreedom:
add toggle ability to NERDTreeCWD.
because NERDTreeToggle changes NERD root to current buffer only at first trigger, then the root will not change again after i change to another buffer, even i set a path argument when i call NERDTreeToggle.NERDTreeCWD does good at change root dir. so we can use NERDTreeCWD instead of NERDTreeToggle.

### DIFF
--- a/plugin/NERD_tree.vim
+++ b/plugin/NERD_tree.vim
@@ -204,8 +204,12 @@ function! NERDTreeFocus()
 endfunction
 
 function! NERDTreeCWD()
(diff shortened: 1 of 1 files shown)


PhilRunninger:
This change would break the behavior untold numbers of people would be expecting of NERDTreeCWD. This is something that would be good to keep in your .vimrc instead.

Issue #650, opened 2016-12-30
set mouse=a


piaoyizhimei:
When i set 'mouse=a'. i get the following error:

处理 function nerdtree#ui_glue#invokeKeyMap[1]..75[50]..74[5]..<SNR>32_handleMiddleMouse 时发生错误:
E117: Unknown function: nerdtree#openExplorer

anhbh:
I got this problem too.

PR #675, opened 2017-03-17
Use relative path with globpath

jason0x43:
Wildignore rules that involve directories aren't handled properly (or at least as one might intuitively expect) when globpath uses absolute paths. For example, when the cwd is `/home/user/project` and wildignore contains `src/*.js`, `globpath("src", "*")` works as expected, while `globpath("/home/user/project/src", "*")` doesn't (js files aren't ignored).

### DIFF
--- a/lib/nerdtree/tree_dir_node.vim
+++ b/lib/nerdtree/tree_dir_node.vim
@@ -272,7 +272,9 @@ function! s:TreeDirNode._initChildren(silent)
 
     "get an array of all the files in the nodes dir
     let dir = self.path
-    let globDir = dir.str({'format': 'Glob'})
+
+    "use a relative globDir so that relative wildignore rules will be obeyed
+    let globDir = fnamemodify(dir.str({'format': 'Glob'}), ':.')
 
     if version >= 703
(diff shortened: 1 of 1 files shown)


lifecrisis:
I'm having real trouble with the NERDTree after this PR. I'm using a Windows machine at work, and applying these commits breaks the `u` mapping. It seems that the NERDTree will no longer allow navigation to directories above the directory in which it was opened.

I'll remain on a test branch until we can resolve this.

jason0x43:
I think I see the issue. The docs for `fnamemodify` state that `:.` will leave a path unmodified if it's not under the cwd. However, in Windows it strips off the drive letter.

lifecrisis:
Thanks, @jason0x43.
I suspected something like this, but I don't have time to confirm or really track it down (I've been working on a fix for issue #650 for the past two days). If you can set up a branch that incorporates your enhancement plus some kind of fix for the problem, I'll be glad to test it on Monday morning.

jason0x43:
Sure, I'll set something up later today or tomorrow.

lifecrisis:
Hey, @jason0x43, thanks for looking into this. I just opened issue #709 for you to close when this problem is definitively resolved.

jason0x43:
I submitted #710 for this. It references #709.

Issue #684, opened 2017-04-20
NERDTree: Refreshing the root node. This could take a while...


gpakosz:
**Environment:**

* Operating System: Mac OS 10.11.6
* Vim version `:version`: VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Apr 11 2017 10:04:07)
* NERDTree version `git rev-parse --short HEAD`: 45f4d61f04e7ef33360f7735931da9ea2ebc05e0
* NERDTree settings applied in your vimrc, if any:

```
" -- NERDTree settings ---------------------------------------------------------

if has("autocmd")
	" open a NERDTree when vim starts up with no files specified
  autocmd StdinReadPre * let s:std_in=1
  autocmd VimEnter * if argc() == 0 && !exists("s:std_in") | NERDTree | endif

  " close vim if the only window left open is a NERDTree
  autocmd BufEnter * if (winnr("$") == 1 && exists("b:NERDTreeType") && b:NERDTreeType == "primary") | q | endif

  autocmd WinEnter * if exists('b:NERDTree') | execute 'normal R' | endif
endif

" toggle NERDTree with <C-N>
nnoremap <silent> <C-N> :NERDTreeToggle<CR>
" find current buffer in NERDTree with <C-F>
nnoremap <silent> <C-F> :NERDTreeFind<CR>

let NERDTreeBookmarksFile=$HOME."/.vim/.NERDTreeBookmarks"
let NERDTreeQuitOnOpen=1
let NERDTreeShowBookmarks=1
let NERDTreeShowHidden=1

set noequalalways " prevents windows size equalization after NERDTree closed
```

**Process:**

1. `$ vim path/to/file`
2. hit <kbd>Ctrl</kbd>+<kbd>f</kbd>
3. navigate to another file with <kbd>Up</kbd>/<kbd>Down</kbd>
4. hit <kbd>Return</kbd>

**Current Result:**

NERDTree displays:
```
NERDTree: Refreshing the root node. This could take a while...
```
and `DONE` is somewhat blinking before it's really done.

**Expected Result:**

Previously, NERDTree would just open the file without trying to refresh its tree.

---
**Optional**

Regression has been introduced by commit `35b04fa3dfdef5cbfbcd6ada59fce3a7bc31eeaa`:
```
commit 35b04fa3dfdef5cbfbcd6ada59fce3a7bc31eeaa
Author: Greg Hurrell <greg@hurrell.net>
[...]

PR #693, opened 2017-05-20
Added back WinEnter and WinLeave to eventignore in nerdtree#exec(cmd)

gpakosz:
NERDTree relies on executing 'wincmd p' back and forth as part of its
normal operation.

This implementation detail is not obvious to NERDTree users that may
set up autocommands.

As such, it makes sense to ignore WinEnter and WinLeave autocmnds in
nerdtree#exec(cmd).


PhilRunninger:
I am going to reject this PR for the reasons discussed in #684.

PR #697, opened 2017-05-29
Key x should also close the DirNode under the cursor

iology:


iology:
Seems that I misunderstand the role of g:NERDTreeCascadeOpenSingleChildDir in this function.

PR #699, opened 2017-06-04
Update docs with already existing mapping variables

asnr:
There are three keymaps already present in the code that aren't included in the docs. I am currently using these mappings (`g:NERDTreeMapPreview`, `g:NERDTreeMapPreviewSplit` and `g:NERDTreeMapPreviewVSplit`) in my .vimrc and can confirm that they work correctly.

This pull requests updates the documentation to mention these mappings.


PhilRunninger:
Good catch @asnr. Thanks for the update.

Issue #677, opened 2017-03-28
MacVim bookmark error 


deltaskelta:
MacVim: Version 8.0.502 (129)
NERDTree: 5.0.0

After making a bookmark of a node I get this error when I am trying to open the bookmark...

```
Error detected while processing function nerdtree#ui_glue#openBookmark[2]..87:
line    2:
E121: Undefined variable: nerdtree
E116: Invalid arguments for function 86
E15: Invalid expression: bookmark.getNode(nerdtree, a:searchFromAbsoluteRoot)

```

lifecrisis:
Hi, @deltaskelta. Thanks for bringing this to our attention.

This looks like a bug that should have surfaced quite some time ago. I have a fix for you to try...

```patch
From 953fa16ffe14cc38c9f558043ca4b238a71d1fda Mon Sep 17 00:00:00 2001
From: Jason Franklin <j_fra@fastmail.us>
Date: Fri, 2 Jun 2017 19:49:51 -0400
Subject: [PATCH] Add a missing function argument sigil

An error message from issue #677 hinted that an argument sigil was
missing in a function in the "bookmark.vim" script. It is unknown why
this bug did not surface until now, but this commit puts the simple fix
in place.

Fixes #677.
---
 lib/nerdtree/bookmark.vim | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/lib/nerdtree/bookmark.vim b/lib/nerdtree/bookmark.vim
index ad8c3d1..e9479ac 100644
--- a/lib/nerdtree/bookmark.vim
+++ b/lib/nerdtree/bookmark.vim
@@ -163,7 +163,7 @@ endfunction
 " treenode for it.
 function! s:Bookmark.GetNodeForName(name, searchFromAbsoluteRoot, nerdtree)
     let bookmark = s:Bookmark.BookmarkFor(a:name)
-    return bookmark.getNode(nerdtree, a:searchFromAbsoluteRoot)
+    return bookmark.getNode(a:nerdtree, a:searchFromAbsoluteRoot)
 endfunction
 
 " FUNCTION: Bookmark.GetSelected() {{{1
-- 
2.7.4

```

Apply the above patch to your local NERDTree repository (since the change is small, you could just do it manually).  If it works for you, I'll submit a pull request to @PhilRunninger to make the change public.

deltaskelta:
[...]

PR #703, opened 2017-06-10
BUGFIX: Repair the broken ":OpenBookmark" command

lifecrisis:
Issue #677 pointed out that several Bookmark commands were  broken by a minor syntax error. The `:OpenBookmark` command was the most obvious among these (which can be used to replicate the error in issue #677). But others included...

`:RevealBookmark`
`:Bookmark` (users couldn't overwrite a Bookmark as the documentation promised)

The latter command was fixed as a side-effect of this PR. The first command is still unstable. The primary purpose of this PR was to fix `:OpenBookmark`, which was accomplished.


PhilRunninger:
Thanks, @lifecrisis. Good work!

PR #705, opened 2017-06-13
Updates FAQ for opening vim on directory

samjonester:
Rearranges the command used to open NERDTree automatically when vim starts up on opening a directory. Now the command will set focus upon opening to the NERDTree split.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -104,7 +104,7 @@ Note: Now start vim with plain `vim`, not `vim .`
 > How can I open NERDTree automatically when vim starts up on opening a directory?
 
     autocmd StdinReadPre * let s:std_in=1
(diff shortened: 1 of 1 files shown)


PhilRunninger:
@samjonester, I view this not as a correction but an alternate, one where focus ends up in the NERDTree window. I'd rather see both options in the FAQ with a sentence that describes the difference between the two.

Thanks for the contribution.

samjonester:
Sure. The biggest thing for me was consistency. The suggested commands when
opening a directory and opening without a file/directory have inconsistent
behavior. With *just* this PR the two commands have consistent behavior.

PhilRunninger:
Good point about the consistency. Another minor point... In your revision, `wincmd p` is no longer needed.

PhilRunninger:
I'm closing this PR. FAQs are being moved to the Wiki, and this new answer has been included there: https://github.com/scrooloose/nerdtree/wiki/F.A.Q.

Issue #709, opened 2017-06-17
Mapping broken by pull request #675


lifecrisis:
Pull request #675 breaks the NERDTree mapping for `u` in instances of Vim running on Windows systems. When `u` is triggered, the parent directory is opened, but all children of the new root node are declared to be invalid and, thus, are not rendered.

I am already in contact with @jason0x43, the author of PR #675, and a fix is in progress.

lifecrisis:
This issue was fixed by PR #710.

PR #710, opened 2017-06-18
Handle non-relative globDir properly in Windows

jason0x43:
In Windows, fnamemodify with ':.' will strip the drive letter from a path that isn't under CWD. This behavior caused a recent update to tree_dir_node to break directory listings in Windows for directories above CWD. The issue was fixed by ensuring that the drive letter is added back to non-relative globDirs on Windows.

references #709

@lifecrisis will probably want to look over this.

### DIFF
--- a/lib/nerdtree/tree_dir_node.vim
+++ b/lib/nerdtree/tree_dir_node.vim
@@ -206,6 +206,25 @@ function! s:TreeDirNode.getDirChildren()
     return filter(self.children, 'v:val.path.isDirectory == 1')
 endfunction
 
+"FUNCTION: TreeDirNode._getGlobDir() {{{1
+"Return a string giving the pathname related to this TreeDirNode. The returned
+"pathname is in a glob-friendly format and is relative to the current working
+"directory, if this TreeDirNode's path is under the current working directory.
+function! s:TreeDirNode._getGlobDir()
+    " Gets a relative path, if possible. This ensures that 'wildignore' rules
+    " for relative paths will be obeyed.
+    let l:globDir = fnamemodify(self.path.str({'format': 'Glob'}), ':.')
+
+    " Calling fnamemodify() with ':.' on Windows systems strips the leading
+    " drive letter from paths that aren't under the working directory. Here,
+    " the drive letter is added back to the pathname.
+    if nerdtree#runningWindows() && l:globDir[0] == '\'
+        let l:globDir = self.path.drive . l:globDir
+    endif
+
+    return l:globDir
+endfunction
+
 "FUNCTION: TreeDirNode.GetSelected() {{{1
 "Returns the current node if it is a dir node, or else returns the current
 "nodes parent
@@ -271,10 +290,7 @@ function! s:TreeDirNode._initChildren(silent)
     let self.children = []
 
     "get an array of all the files in the nodes dir
-    let dir = self.path
-
-    "use a relative globDir so that relative wildignore rules will be obeyed
-    let globDir = fnamemodify(dir.str({'format': 'Glob'}), ':.')
+    let globDir = self._getGlobDir()
 
     if version >= 703
         let filesStr = globpath(globDir, '*', !g:NERDTreeRespectWildIgnore) . "\n" . globpath(globDir, '.*', !g:NERDTreeRespectWildIgnore)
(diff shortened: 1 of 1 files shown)


lifecrisis:
Thanks again, @jason0x43. I'll take a look in the morning and report back.

lifecrisis:
Two comments:

1. I disagree with adding a new class method here. There is only one block to add: the conditional for tweaking the filename modification on Windows systems. That change is part of the logic of the two methods and should be completely contained in those methods. In addition, the namespace of functions in the various NERDTree classes is already quite cluttered and doesn't need any additions without a very good reason. If we need to use this code a third or fourth time elsewhere, then we should consider making a new method.

2. Your function name also does not obey proper NERDTree naming conventions (see the blog post mentioned in `:h NERDTree` for a discussion of these naming rules).

I hope that you and @PhilRunninger will at least consider my suggestions. Other than that, the code doesn't break my configuration anymore, so I am okay with the change and its intent. I simply think its footprint in the code could be much more compact.

**P.S.**  It's worth pointing out that `fnamemodify()` is misbehaving here... this is a bug with a Vim builtin. At least, the behavior of the builtin does not agree with the documentation:

```
	:.	Reduce file name to be relative to current directory, if
		possible.  File name is unmodified if it is not below the
		current directory.
```

This is plain wrong... the drive letter is stripped.

jason0x43:
I'm fine with updating the PR. Just to make sure we're on the same page:

1. I had started out making the necessary change in the two relevant places in tree_dir_node, but it seemed unnecessarily repetitive (and more difficult to maintain) to add the same logic in two different places. How about a script-local, but non-class method?
2. It follows the naming convention of an instance method (so far as I can tell), which was the intention. However, after further consideration, it probably should've been static or private (or a non-member function).

lifecrisis:
If you want to go with a method, I would suggest a private instance method. In order for a method to be considered an instance method, it must reference `self`. Otherwise, it is a class method.

I would suggest a function like the following:

```viml
"FUNCTION: TreeDirNode.getGlobDir(path) {{{1
"Return a string giving the pathname related to this TreeDirNode. The returned
"pathname is in a glob-friendly format and is relative to the current working
"directory, if possible.
function! s:TreeDirNode._getGlobDir()

    " Gets a relative path, if possible. This ensures that 'wildignore' rules
    " for relative paths will be obeyed.
    let l:globDir = fnamemodify(self.path.str({'format': 'Glob'}), ':.')

    " Calling fnamemodify() with ':.' on Windows systems strips the leading
    " drive letter from paths that aren't under the working directory. Here,
    " the drive letter is added back to the pathname.
    if nerdtree#runningWindows() && l:globDir[0] == '\'
        let l:globDir = self.path.drive . l:globDir
    endif

    return l:globDir
endfunction
```

Then you could get rid of the lines like this:

```viml
let dir = self.path
```

Thanks! Also, notice that I used sigils wherever possible. Making these explicit is generally a good idea. I'd like that to remain in the final code.

lifecrisis:
After another look, the above is definitely all the input I have, thanks for the contribution!

jason0x43:
Makes sense. I updated the PR to use a private method and removed the redundant `self.path` accesses.

That's interesting about class vs instance methods (that the differentiator is whether a method accesses self, regardless of whether it _could_ access self). I've never actually dealt with OOP in vim.

lifecrisis:
Thanks!

I think the distinction is less of a rule than a convention. In fact, *all* dictionary methods can access `self` provided they were defined in the standard way (this means that even our "class methods" could access `self`).  So, the division is really one of style.  Following it closely makes the code much better.

See this [blog post](http://got-ravings.blogspot.com/2008/09/vim-pr0n-prototype-based-objects.html) if you're interested in what @scrooloose has to say about these ideas. I think I'm following him correctly... The section "Class methods and class properties" addresses this.

PhilRunninger:
Thanks, @lifecrisis  and @jason0x43  for finding and fixing this bug. Great work.

lifecrisis:
By the way, @jason0x43, if you include text like `Fixes #709` in your commit message, the issue will automatically be closed when the commit is merged. That can be pretty useful feature of Github.

jason0x43:
I added a 'references' line so the commit would auto-link to the issue, but for some reason I don't generally use 'fixes' in PRs. It would make life easier, though. 😄

PR #713, opened 2017-06-21
BUGFIX: Compare function failed when sorting bookmarks.

bravestarr:
A class method doesn't match the sort compare function, so use a
plain one instead.

### DIFF
--- a/lib/nerdtree/bookmark.vim
+++ b/lib/nerdtree/bookmark.vim
@@ -117,11 +117,11 @@ function! s:Bookmark.CacheBookmarks(silent)
     endif
 endfunction
 
-" FUNCTION: Bookmark.CompareBookmarksByName(firstBookmark, secondBookmark) {{{1
-" Class method that indicates the relative position of two bookmarks when
+" FUNCTION: CompareBookmarksByName(firstBookmark, secondBookmark) {{{1
+" Function that indicates the relative position of two bookmarks when
 " placed in alphabetical order by name. Case-sensitivity is determined by an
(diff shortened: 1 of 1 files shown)


lifecrisis:
@PhilRunninger, since I implemented this function, I'd like to have some input before a merge.

This issue has been raised before. What version of Vim are you using? The last user solved their problem by updating Vim.

Also, the sort function *can* use a dictionary function, but Vim's treatment of such functions with the `sort()` builtin appears to be inconsistent across versions 7 and 8. Please test your installation with the following line in place of line 277 in `bookmark.vim`:

```viml
call sort(s:Bookmark.Bookmarks(), s:Bookmark.CompareBookmarksByName, s:Bookmark)
```

This may solve your problem, while maintaining the desired object-oriented style and avoiding adding another function to the global namespace.

bravestarr:
Yes, my Vim is too old (exact version 7.4).  I try your solution, appending s:Bookmark as the third argument for sort, it works well.  Then I update my Vim to the latest (version 8.0.586), both the ways - with or without  dict for sort, work fine.  Thank you for your quick response!

lifecrisis:
No problem! So @bravestarr, just to confirm, adding the fix I posted above worked *before* you updated Vim? Is that the case?

It would be nice to know if that fix worked for an older version of Vim that have this problem. It doesn't seem to break later versions. Thus, it might be worth including that small change (the new argument) so that this issue doesn't come up again.

bravestarr:
Your fix is confirmed to work under v7.4 Vim.  Also it works under the latest version (v8.0.586), so I think you may consider to commit it for backward compatibility.

PR #717, opened 2017-06-28
Define the "TreeDirNode._glob()" method.

lifecrisis:
I think my methods for managing this refactor (in terms of committing the changes here) was too complicated.

I experimented in the past two weeks with different methods for adding commits. I used "squash and merge" as well as "rebase and merge". They did work, but they were a little complicated to get right. I'm still somewhat new to using Git, and best practices for administering a repository haven't yet registered with me.

Nothing is wrong with the new techniques, but they seem much better for cleaning up a repository rather than adding new commits. In the future, I'll stick with the practice of adding regular old merge commits built from simple topic branches. This makes it easier for pretty much anyone to follow the development of the project.

@PhilRunninger, is going back to the usual practice preferable to you? It makes sense to me...

### DIFF
--- a/lib/nerdtree/tree_dir_node.vim
+++ b/lib/nerdtree/tree_dir_node.vim
@@ -206,15 +206,23 @@ function! s:TreeDirNode.getDirChildren()
     return filter(self.children, 'v:val.path.isDirectory == 1')
 endfunction
 
-" FUNCTION: TreeDirNode._getGlobDir() {{{1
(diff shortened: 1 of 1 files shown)


PhilRunninger:
FWIW, I don't know that much about rebasing, and I have never seen a real need to use it. I used squash once, and ended up putting five commits' messages in the one squashed commit. Squashing loses the granularity you see with multiple commits. Sometimes it's nice to see the incremental changes that multiple commits will show you.

PR #720, opened 2017-07-02
BUGFIX: Repair opening and closing of cascades.

lifecrisis:
This pull request fixes a common complaint with the cascade feature of the NERDTree. See the commit messages for details.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -59,7 +59,7 @@ Installation
 
     git clone https://github.com/scrooloose/nerdtree.git ~/.vim/bundle/nerdtree
 
-Then reload vim, run `:helptags ~/.vim/bundle/nerdtree/doc/`, and check out `:help NERD_tree.txt`.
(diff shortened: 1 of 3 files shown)


lifecrisis:
Currently, we cannot remove that option.

My understanding is that `g:NERDTreeCascadeOpenSingleChildDir` affects the *behaviour* of a cascade (i.e., it opens as many directories as possible), whereas `g:NERDTreeCascadeSingleChildDir` affects the *appearance* (i.e., it says to render a cascade on a single line). This isn't very clear from the long and confusing names... I was only able to discern this by reading the code directly.

If this is the difference, however, it would be useful to keep both options to allow granular control. Renaming and re-documenting these settings is advisable.

lifecrisis:
By the way, issue #519 is also more thoroughly fixed by this pull request, the workaround you mentioned is no longer needed.

lifecrisis:
@PhilRunninger, I'm going to let this sit for a few more days and then merge when no complaints come in. I'm signing off now though. Enjoy your 4th of July!

PR #724, opened 2017-07-19
1,add a submenu open file use gnome-open or xdg-open

rowanpang:
tks for your feedback @lifecrisis and @PhilRunninger . On your recommend  I have made a patch let the `gx` do the work.

### DIFF
--- a/nerdtree_plugin/menu_sys_open.vim
+++ b/nerdtree_plugin/menu_sys_open.vim
@@ -14,7 +14,14 @@ function! s:callback_name()
 endfunction
 
 function! s:callback()
-    if exists('*vimproc#open')
+    if !exists("g:netrw_nogx") 
+	let oldpwd = getcwd()
+        let ret = g:NERDTreeFileNode.GetSelected().path.changeToDir()
+	if !ret
+	    exe "normal gx"
+	    exe "cd" . oldpwd 
+	endif
+    elseif exists('*vimproc#open')
         call vimproc#open(g:NERDTreeFileNode.GetSelected().path.str())
     else
         let path = g:NERDTreeFileNode.GetSelected().path.str({'escape': 1})
@@ -24,8 +31,10 @@ function! s:callback()
             echoerr "or install vimproc from 'https://github.com/Shougo/vimproc'"
             return
         endif
-        let cmd = g:nerdtree_plugin_open_cmd . " " . path
-        call system(cmd)
+
+	let cmd = g:nerdtree_plugin_open_cmd . " " . path
+	call system(cmd)
+
     endif
 endfunction
 
--- a/nerdtree_plugin/menu_sys_open.vim
+++ b/nerdtree_plugin/menu_sys_open.vim
@@ -1,0 +1,42 @@
+" ============================================================================
+" File:        menu_sys_open.vim
+" Description: plugin for NERD Tree that provides an menu item for open file
+" License:     This program is free software. It comes without any warranty,
+"              to the extent permitted by applicable law. 
+"
+if exists("g:loaded_nerdtree_plugin_open")
+    finish
+endif
+let g:loaded_nerdtree_plugin_open = 1
+
+function! s:callback_name()
+    return matchstr(expand('<sfile>'), '<SNR>\d\+_') . 'callback'
+endfunction
+
+function! s:callback()
+    if exists('*vimproc#open')
+        call vimproc#open(g:NERDTreeFileNode.GetSelected().path.str())
+    else
+        let path = g:NERDTreeFileNode.GetSelected().path.str({'escape': 1})
+
+        if !exists("g:nerdtree_plugin_open_cmd")
+            echoerr "please set 'g:nerdtree_plugin_open_cmd'  to 'open','gnome-open' or 'xdg-open'"
+            echoerr "or install vimproc from 'https://github.com/Shougo/vimproc'"
+            return
+        endif
+        let cmd = g:nerdtree_plugin_open_cmd . " " . path
+        call system(cmd)
+    endif
+endfunction
+
+call NERDTreeAddKeyMap({
+	\ 'callback': s:callback_name(),
+	\ 'quickhelpText': 'direct call sys open',
+	\ 'key': 'E',
+	\ })
+
+call NERDTreeAddMenuItem({
+	\ 'text': '(o)open with system command',
+	\ 'shortcut': 'o',
+	\ 'callback': s:callback_name()
+	\ })


lifecrisis:
I'm inclined to reject this PR, but I'd like some feedback from @PhilRunninger.

Vim already ships with a macro for opening files in this manner. That macro is `gx`.  It already works on Windows, KDE, Mac OS X, and Gnome. Further, it can be extended easily to work on other platforms (by defining a few variables in your `.vimrc` file.

It is better to rely on this macro since it comes with vim and requires no special work. To use the macro in the NERDTree, execute `cd` followed by `gx`.

A better feature-addition to the NERDTree would be to create a wrapper around `gx` so that this well-supported and well-tested functionality does not go to waste.

PhilRunninger:
`open` is already supported by NERDTree's `m` `o` commmand, [here](https://github.com/scrooloose/nerdtree/blob/master/nerdtree_plugin/fs_menu.vim#L280-L287). So now the question becomes: Do we

1. make it configurable like this PR does,
1. use the `has()` function to determine which `open` command to use, or
1. or rewrite the function to use `gx`?

Of these, I like the 3rd option, although I don't like the possible side effect of it doing a `cd` then `gx`. The function would need to restore the original working directory. I concur with rejecting the PR, and would accept one implementing the `gx` wrapper in place of the current function.

lifecrisis:
The `gx` mapping would also be useful as a normal mapping for the NERDTree. This could be done through the API, however.

lifecrisis:
There might be an issue with the behaviour of the `gx` macro in that it prompts for a `<CR>` before returning focus to Vim as you would expect. I've looked in the `runtime/` directory of my Vim distribution, and this seems to be built into the function that drives `gx`.

This is only on Windows, however. On my Linux machine, no prompt for `<CR>` is given.

lifecrisis:
The code at this point looks okay. Thanks for making the requested changes @rowanpang. However, I'd like to make a few commits that will improve its style. I will review further this weekend.

@PhilRunninger and @rowanpang, are you both okay with this?

nerdtree_plugin/menu_sys_open.vim
```diff
@@ -14,7 +14,14 @@ function! s:callback_name()
...
 function! s:callback()
-    if exists('*vimproc#open')
+    if !exists("g:netrw_nogx") 
+	let oldpwd = getcwd()
+        let ret = g:NERDTreeFileNode.GetSelected().path.changeToDir()
+	if !ret
+	    exe "normal gx"
+	    exe "cd" . oldpwd 
+	endif
+    elseif exists('*vimproc#open')
```
PhilRunninger: My only concern here is that NERDTree would be tightly coupled to another plugin, over which we have no control. I think this coupling is best handled in a user's `.vimrc` file.

lifecrisis:
@PhilRunninger, this comment also applies to your observation above.

There are several problems here, mainly because this is a fundamentally complex thing to do.

I'm not worried about using the mapping for `<Plug>NetrwBrowseX`... that is a public feature that is distributed with Vim, so it will likely be very stable. It's more of an issue of getting the behaviour of our mapping to match what NERDTree users would expect. The netrw plugin exposes a nice function that will do this, and I'm confident that it's dependable (provided we check that netrw is enabled).

The following will do the job in a clean way:

```viml
call netrw#BrowseX(g:NERDTreeFileNode.GetSelected().path.str(), 0)
```

(By the way, the netrw code is ugly as hell.)

In the NERDTree, position your cursor on any file (your current working directory must be its parent). The above line will effectively trigger `gx` in a nicer way. So, all a user has to do is use this call in a custom mapping, and the desired functionality is there.

@rowanpang, I suggest that you write your own custom mapping using the map API that calls the function here. That will give you the functionality you need. A feature like this should be implemented at some point, but I vote that that be made a bona fide project for the next version. I apologize for the long discussion, and appreciate your work. It helped us get a lot of these ideas out in the open.

@PhilRunninger, do you agree?

lifecrisis:
To further support you @rowanpang, I've written the following mapping:

```viml
function! g:OpenFileHandler(fileNode)
    call g:netrw#BrowseX(a:fileNode.path.str(), 0)
endfunction

call g:NERDTreeAddKeyMap({
            \ 'key' : 'gx',
            \ 'callback' : 'OpenFileHandler',
            \ 'quickhelpText' : 'open with external program',
            \ 'scope' : 'FileNode',
            \ })
```

You can put this in your `nerdtree_plugin/` directory and get the desired functionality. If you want to open an explorer for directory nodes, this function can be easily modified to do that.

lifecrisis:
**UPDATE:**

I just got an email from Chip Campbell, author of netrw. He's removing the `press <cr>` prompts from netrw on Windows. I investigated the problem, and it turned out they weren't necessary in Windows 10.

I'm still on the fence here. The problem is nearly solved without this pull request, but I'm not ready to close yet. I'll review this again later.

lifecrisis:
Okay, so I've decided to close this PR without merging.

The main reason is that we've established these changes can be made in your `nerdtree_plugin/` directory. You can add custom maps *and* menu items there, rather than in the NERDTree codebase itself.

Any feature additions that can be added within the `nerdtree_plugin/` directory should be added there. Otherwise, the NERDTree codebase would quickly grow beyond control.

I still believe this was productive. Thanks to @rowanpang.

thiswillbeyourgithub:
Hi,
Ssorry to unbury this very old thread but I have been using `gx` as for some inexplicable reasons `xdg-open` doesn't work with pdfs inside nerdtree (it works perfectly fine in any other terminal, if anyone has an idea I'm all ears).

I'm here because `gx` seems to do the job fine but nerdtree is unresponsive as long as the app is open, and resumes when the app is closed. Is there a way to detach the app from nvim when it is launched ?

Thanks !

NERDTree version 6.6.1
nvim version 0.4.3
ubuntu 18.04 with i3wm as window manager

hyhugh:
> Hi,
> Ssorry to unbury this very old thread but I have been using `gx` as for some inexplicable reasons `xdg-open` doesn't work with pdfs inside nerdtree (it works perfectly fine in any other terminal, if anyone has an idea I'm all ears).
> 
> I'm here because `gx` (as described [here](https://github.com/preservim/nerdtree/pull/724#issuecomment-316986478)) seems to do the job fine but nerdtree becomes unresponsive as long as the app is open, and resumes when the app is closed. Is there a way to detach the app from nvim when it is launched ?
> 
> Thanks !
> 
> NERDTree version 6.6.1
> nvim version 0.4.3
> ubuntu 18.04 with i3wm as window manager

Hi, 
Please set g:netrw_nogx in your .vimrc and then nerdtree will stop using gx to open files.

```
let g:netrw_nogx = 1
```

hyhugh:
> Hi,
> Ssorry to unbury this very old thread but I have been using `gx` as for some inexplicable reasons `xdg-open` doesn't work with pdfs inside nerdtree (it works perfectly fine in any other terminal, if anyone has an idea I'm all ears).
> 
> I'm here because `gx` (as described [here](https://github.com/preservim/nerdtree/pull/724#issuecomment-316986478)) seems to do the job fine but nerdtree becomes unresponsive as long as the app is open, and resumes when the app is closed. Is there a way to detach the app from nvim when it is launched ?
> 
> Thanks !
> 
> NERDTree version 6.6.1
> nvim version 0.4.3
> ubuntu 18.04 with i3wm as window manager

Hi.
I just created a plugin to solve this issue
please check
https://github.com/hyhugh/nerdtree-async-xdg-open.vim
after installing it, you can use the newly added item to open file asynchronously

PR #729, opened 2017-07-24
Use window number when opening in new vsplit

ParkerKemp:
Currently, NERDTree uses "wincmd p" to switch between windows while creating a new vsplit. 

This behavior causes an issue when used in conjunction with the golden-ratio plugin, because (presumably) that plugin handles the window resize event, and in doing so changes what the "previous window" is. As a result, NERDTree then opens the selected file in the wrong window, replacing the existing buffer and leaving a blank vertical window adjacent to it.

This very small commit tweaks the _newVSplit() function to get the actual window number of the new vsplit and return to it after resizing the tree window.

### DIFF
--- a/lib/nerdtree/opener.vim
+++ b/lib/nerdtree/opener.vim
@@ -225,10 +225,12 @@ function! s:Opener._newVSplit()
     call nerdtree#exec("wincmd p")
     vnew
 
+    let currentWin = winnr()
+
     "resize the nerd tree back to the original size
     call g:NERDTree.CursorToTreeWin()
     exec("silent vertical resize ". winwidth)
-    call nerdtree#exec('wincmd p')
+    exe currentWin . "wincmd w"
 endfunction
 
 "FUNCTION: Opener.open(target) {{{1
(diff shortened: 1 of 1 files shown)


lifecrisis:
This is a simple enough change. I might want to do one or two small style things here...

You didn't use `call nerdtree#exec(currentWindow . 'wincmd w')` in your change. Is there any particular reason why?

ParkerKemp:
No reason. I'm new to this codebase and I guess I overlooked some of the code style. Sorry about that.

PhilRunninger:
Does a similar edit need to be done in the `Opener._newSplit()` function just above this one?

lifecrisis:
Hey @ParkerKemp, this all looks good. I'll make some additions this weekend. Otherwise, consider this accepted.

ParkerKemp:
Sounds great, glad I was able to contribute. Let me know if you need anything from me.

lifecrisis:
**UPDATE:**

I've just finished my edits. I didn't change `_newSplit()` because there were no complaints for this method. In addition, that function is complex and needs special attention in a separate PR.

Thanks to @ParkerKemp.

PR #734, opened 2017-08-11
BUGFIX: Make the NERDTree aware of 'shellslash'

lifecrisis:
This PR fixes several issues (5 exactly) related to editing files listed in the NERDTree while `'shellslash'` is set on Windows systems.

Before this PR, users could not `cd` to a directory with a '$' in the name and could not open files with parentheses or spaces in the name (again, on Windows while using 'shellslash').

Most of this was related to neglecting to use the helpful `fnameescape()` function included with Vim.

**NOTE:** This is a fairly major change and testing/commentary/discussion is needed!!!

### DIFF
--- a/lib/nerdtree/path.vim
+++ b/lib/nerdtree/path.vim
@@ -640,27 +640,27 @@ function! s:Path.rename(newPath)
 endfunction
 
 " FUNCTION: Path.str() {{{1
+" Return a string representation of this Path object.
 "
-" Returns a string representation of this Path
+" Args:
+" This function takes a single dictionary (optional) with keys and values that
+" specify how the returned pathname should be formatted.
 "
-" Takes an optional dictionary param to specify how the output should be
-" formatted.
-"
-" The dict may have the following keys:
+" The dictionary may have the following keys:
 "  'format'
 "  'escape'
 "  'truncateTo'
 "
 " The 'format' key may have a value of:
-"  'Cd' - a string to be used with the :cd command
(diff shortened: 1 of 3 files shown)


PhilRunninger:
I may have found another issue: the `m` `l` menu option isn't working when `'shellslash'` is on. I get the following error messages:
```
Error detected while processing function nerdtree#ui_glue#invokeKeyMap[1]..74[33]..73[3]..<SNR>44_showMenu[2]..48[19]..65[6]..NERDTreeListNodeWin32:
line    4:
E684: list index out of range: 0
E116: Invalid arguments for function nerdtree#echo
```

lifecrisis:
Thanks, @PhilRunninger. We'll leave this here until I can address that. I'm in no rush!

PhilRunninger:
The `m` `l` bug was **not** introduced by this PR, but there may (probably) be a line of code to be changed that the changes in this PR would fix.

lifecrisis:
Hey, @PhilRunninger, that problem you mentioned seems to only be an issue when `'shellslash'` is set.  it looks like the menu command you're using calls a Windows shell command. When you use `'shellslash'`, the file that is serving as an operand will contain forward slashes and thus cause the command to fail (because the shell command is meant for `cmd.exe`).

This can be fixed fairly easily, but I'm not feeling like working this weekend. I'll get to it eventually.

lifecrisis:
Hey @PhilRunninger, I had a burst of motivation this morning and went ahead and fixed that problem. Everything looks good so I'll go ahead and merge.

PR #739, opened 2017-09-14
Add a 'NERDTreeToggleHere' command

mmequignon:
Targets the current file with the toggle functionnality.

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -246,6 +246,15 @@ function! s:displayHelp()
     call b:NERDTree.ui.centerView()
 endfunction
 
+" FUNCTION: s:ToggleTabHere() {{{1
+function! s:ToggleTabHere()
+    if g:NERDTree.IsOpen()
+        call g:NERDTree.Close()
+    else
+        call s:findAndRevealPath()
+    endif
+endfunction
+
 " FUNCTION: s:findAndRevealPath() {{{1
 function! s:findAndRevealPath()
     try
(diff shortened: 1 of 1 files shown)


lifecrisis:
This is a nice idea, but your commit doesn't really contribute to a core feature of the NERDTree.  This type of customization could just as easily have been placed in your `vimrc` file.  Thanks for the effort, but I'll have to motion that this PR be closed.

I'll wait on my colleague @PhilRunninger to weigh in before closing.

mmequignon:
Yes, you're right.
But I find that implementing this "feature" in the .vimrc is not that easy.
I tried to use the NERDTreeToggle optionnal path dynamically. (something like `:NERDTreeToggle expand('%:p')`) but it didn't worked.
I tried to implement the code of my commit in my .virmrc, but I didn't succeeded either.
Maybe it would be interesting to explain in the README how to expand default NERDTree behaviours ?

lifecrisis:
The reason your command didn't work is because you can't pass an expression as an argument to the `:NERDTreeToggle` ex command.

You need to use the `:execute` command to make it work. See the snippet below...

```viml
:execute 'NERDTreeToggle ' . expand('%:p')
```

That line will function as you expect.

lifecrisis:
It is worth noting that `:NERDTreeToggle` does *not* obey an argument when a NERDTree already exists for the current tab.  It only uses its argument on the first invocation. This might be considered a bug by some. But it can easily be worked around for now.

lifecrisis:
As a final note, you could modify your function like so to include it within your `vimrc` file:

```viml
function! s:ToggleTabHere()
    if g:NERDTree.IsOpen()
        call g:NERDTree.Close()
    else
        NERDTreeFind
    endif
endfunction
```

Then, just create a map or command as you wish for your personal use.

mmequignon:
Thank you for your help.
I close the pr.

PhilRunninger:
@mmequignon , here's another, perhaps simpler, option for you. Try mapping this expression to a key in your .vimrc.
```
nnoremap <silent><expr> <leader>f g:NERDTree.IsOpen() ? ":NERDTreeClose\<CR>" : ":NERDTreeFind\<CR>"
```

mmequignon:
Ternary operator, yes.
Thank you !

PR #743, opened 2017-09-24
echo a message to alert the user when change tree root using bookmarks or press 'C'(when  g:NERDTreeChDirMode==#2)

skyblueee:
when g:NERDTreeChDirMode==#2 and
1. :BookmarkOpen <bm> OR press 'o' on bookmarks
2. press 'C' to change tree root to the selected dir
make cwd autochange and give hints to users( just as the same with u/U)

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -141,6 +141,9 @@ endfunction
 " changes the current root to the selected one
 function! s:chRoot(node)
     call b:NERDTree.changeRoot(a:node)
+    if g:NERDTreeChDirMode ==# 2
+        call b:NERDTree.root.path.changeToDir()
+    endif
 endfunction
 
 " FUNCTION: s:nerdtree#ui_glue#chRootCwd() {{{1
@@ -474,6 +477,10 @@ function! nerdtree#ui_glue#openBookmark(name)
         call l:bookmark.open(b:NERDTree)
     else
         call l:bookmark.open(b:NERDTree, {'where': 'p'})
+    endif
+
+    if g:NERDTreeChDirMode ==# 2
+        call b:NERDTree.root.path.changeToDir()
     endif
 endfunction
 
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -141,9 +141,6 @@ endfunction
 " changes the current root to the selected one
 function! s:chRoot(node)
     call b:NERDTree.changeRoot(a:node)
-    if g:NERDTreeChDirMode ==# 2
-        call b:NERDTree.root.path.changeToDir()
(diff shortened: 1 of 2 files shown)


lifecrisis:
I can't replicate the bugs that your pull request purports to fix.  The `NERDTree.changeRoot(node)` function always performs a check of the `g:NERDTreeChDirMode` option and behaves accordingly.  Thus, your changes in `ui_glue.vim` are redundant... they are also examples of code duplication that should be avoided (i.e., both of those functions eventually call `changeRoot()`).

However, the change you made in `nerdtree.vim` is valid. It calls a better function to change the directory that also echoes a message to alert the user.  I like this change and would like to include it.

skyblueee:
Oh, you are right. I'm so sorry for my carelessness and thanks for your reply.

I tested it again to make sure you are right. And now the problem only in nerdtree.vim. And of course it's not called a 'fixing bug', it just make the plugin more friendly.

I have commit it again. :-)

lifecrisis:
Thanks for the prompt reply!

Please fix the indentation of your code, and then I'll be satisfied. We'll wait for approval from @PhilRunninger before merging.  Thanks again!

skyblueee:
Because of my 'noexpandtab' setting, I didn't notice this problem before. 
Have fixed it right now.

skyblueee:
I think it's OK now. And now what should I do? Just wait for a merge?
Actually, I am new in github and wondering what to do at the Deploy step, which is mentioned in https://guides.github.com/introduction/flow/
@lifecrisis @PhilRunninger

PhilRunninger:
The Deploy stage is where someone would check out your branch and make sure NERDTree still works properly. This ensures you've included everything in your commits. Once past that phase, the pull request would be merged into production.

lifecrisis:
Thanks, @PhilRunninger, I'll review one more time and merge tomorrow.

skyblueee:
Thank you both. I find a can click the "close and comment" button now, should I do it now?

Another question, it seems that I have found other things need to be changed and to open a new pull request.
What should I do? Wait for the merge, sync my fork, change, commit, and then open a new pull request? Or just change my fork, commit and open a pull request before this one is closed and merged?

lifecrisis:
At this point you don't have to do anything.  I'll merge the pull request, and the work will be done.

Generally, a new pull request would require a new branch to be created in your fork.  You would then make commits to this branch and open a pull request to have your branch merged into the `master` branch in this repository.  After merging, you can then delete your branch.

Sounds like you need to brush up on your Git skills.  I recommend **Pro Git** by Scott Chacon. It's a free online book.

skyblueee:
@lifecrisis Thanks for your advice very very much! During the past 20+ days I have read over Progit2, and it is very helpful. Now I have corrected the commit history of my fork repo. Thanks again.

PR #744, opened 2017-09-24
mirror a tree which tab already closed

faster-mo:
I am using tabs for multi-file editing instead of buffers, And nerdtree monitor is quite useful for me, But  nerdtree can not monitor a tree which tab closed, So I did it.

### DIFF
--- a/lib/nerdtree/creator.vim
+++ b/lib/nerdtree/creator.vim
@@ -134,6 +134,13 @@ function! s:Creator.createMirror()
             call add(treeBufNames, nextName)
         endif
     endfor
+
+    for i in range(1, bufnr('$'))
+        let nextName = bufname(i)
+        if bufexists(i) && getbufvar(i, '&filetype') == 'nerdtree' && (!exists("t:NERDTreeBufName") || nextName != t:NERDTreeBufName)
+            call add(treeBufNames, nextName)
+        endif
+    endfor
     let treeBufNames = self._uniq(treeBufNames)
 
     "map the option names (that the user will be prompted with) to the nerd


lifecrisis:
The problem that your pull request "solves" is not clear to me.  Based on the short description you provided, I have come to the conclusion that you are using Vim incorrectly.  See [this link](https://stackoverflow.com/questions/26708822/why-do-vim-experts-prefer-buffers-over-tabs#26710166) for a clarification of my position.

I am inclined to reject this pull request.  I'll allow for more discussion before closing.

faster-mo:
I think there is no wrong or right for users, I chose the tab way, tab as file-proxies, vim instance as project. And as an plugin should compatible with all the "way".

lifecrisis:
Well, that still doesn't clarify the intent of your change.  It looks to me like you have introduced redundant code.

To make sure I'm clear, the intent of your change is to have `:NERDTreeMirror` allow the user to select not only from NERDTree instances that currently exist in tabs, but also from those that existed in tabs that have been closed.  Is that a correct summary?

faster-mo:
Yes, You summary is correct.

lifecrisis:
If we were to include this, more changes would be required.  Documentation would need to be updated, and other regions of code would needed to be removed to fix the new redundancy.

More importantly, the design choice here is problematic.  Over the course of an editing session (throughout a typical workday), I may open hundreds of NERDTree buffers.  Your change becomes a problem when these buffers accumulate.  It is my opinion that only active NERDTree buffers should be presented in this menu.

Thanks for your suggestion, but I must suggest that this pull request be closed. I will wait until @PhilRunninger has reviewed the code to make a final decision.

faster-mo:
I want to know how many unique NERDTree buffers when you end up your typical coding work? If you use :NERDTreeMirror instead of :NERDTreeToggle it show be reduce quite a lot. Generally :NERDTreeMirror is  faster than :NERDTreeToggle. 

After decision I will closed this pull request and review my to fixed the redundancy. Then I will make a pull request again.

faster-mo:
Thanks, I got many form the decision.

PR #748, opened 2017-10-16
make nerdtree able to restore window size automatically when toggled …

howmuchsnow:
…on next time after resized


lifecrisis:
Another global variable is not something that the NERDTree should include at this point.  The functionality you're looking for can be achieved with a custom function that wraps the `:NERDTreeToggle` command.

@PhilRunninger, I suggest that this PR be rejected.  I'll be glad to discuss if other maintainers disagree.

PhilRunninger:
I agree.

PR #751, opened 2017-10-27
fixed: directory navigation does not work on Cygwin

ashumkin:
Directory tree navigation is broken because of directory signs which is
shown in UTF-8.

Cygwin is a Windows application, so it uses ASCII codepages and so
directory signs must be in ASCII, but if to modify
"nerdtree#runningWindows" function there are many other functions break
that convert paths, e.g.

So, the quick and reliable solution is to add a
separate function "nerdtree#runningCygwin" and use it in a specific
place.

### DIFF
--- a/autoload/nerdtree.vim
+++ b/autoload/nerdtree.vim
@@ -155,6 +155,11 @@ function! nerdtree#runningWindows()
     return has("win16") || has("win32") || has("win64")
 endfunction
 
+"FUNCTION: nerdtree#runningCygwin(dir) {{{2
+function! nerdtree#runningCygwin()
+    return has("win32unix")
+endfunction
+
 " SECTION: View Functions {{{1
 "============================================================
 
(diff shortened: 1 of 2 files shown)


lifecrisis:
Hmmm, I don't know much about Cygwin, but this change seems okay to me.  Is the `+win32unix` setting always guaranteed to be enabled when Cygwin is in use? @PhilRunninger, do you know anything about this?

PhilRunninger:
This looks like a good solution, although I have no way to test it. I had tried to solve this #602 before (with #626). However it caused other things to break, so I had to revert it #695. Your solution looks like it's the proper fix. @lifecrisis , what do you think?

ashumkin:
> Is the +win32unix setting always guaranteed to be enabled when Cygwin is in use? 

Accoriding to [vim sources](https://github.com/vim/vim/blob/816968defc8ae79eb7e2319e991e74661be8d750/src/evalfunc.c#L5563)  the answer is "_YES, it is guaranteed_"

lifecrisis:
@ashumkin, excellent answer!  You get a thumbs-up for being a true programmer.

@PhilRunninger, I also have no way to test it, as I do not run Cygwin on any of my machines.  However, @ashumkin just gave us an authoritative reason to believe this is the way to go.

I suggest a merge.

PR #753, opened 2017-10-30
Add to FAQ: switch between windows

miparnisari:
This SO question: https://stackoverflow.com/questions/1656591/how-to-jump-back-to-nerdtree-from-file-in-tab has over 170 votes, so I guess you can call this a frequently asked question :) It's also the 3rd most voted question: https://stackoverflow.com/questions/tagged/nerdtree?sort=votes&pageSize=50

In fact, as a new `vim` user, this was the first question that came to my mind when I installed nerdtree.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -31,6 +31,10 @@ Then reload Vim, run `:helptags ~/.vim/bundle/nerdtree/doc/`, and check out `:he
 F.A.Q.
 ------
 
+> How do I switch between the different windows?
+
+Press `Ctrl+ww`.
+
 > Is there any support for `git` flags?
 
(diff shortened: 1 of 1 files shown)


lifecrisis:
This is a useful suggestion, but I don't want to merge it into the `README`.

We should really be moving this kind of documentation into the repository wiki pages.  If you want to start a "F.A.Q." page, that would be a welcome contribution.

miparnisari:
Ok, done: https://github.com/scrooloose/nerdtree/wiki/F.A.Q.

Perhaps we can update the README to point to this link?

lifecrisis:
We'll update the README when the F.A.Q. has been moved completely to the Wiki. In the future, try to use the same notation that the Vim documentation uses for special keys.

miparnisari:
@lifecrisis ... ? i can do it

PR #754, opened 2017-10-30
Display file size with commas as thousands separators.

PhilRunninger:
This change will display the file size in an easier to read format, and
still display its exact value. It uses the stat command (with different
switches and formats between Unix and OSX) to get the file size, sed to
add the commas to it, and sed again to replace the original size with
the modified number.

### DIFF
--- a/nerdtree_plugin/fs_menu.vim
+++ b/nerdtree_plugin/fs_menu.vim
@@ -213,7 +213,7 @@ function! NERDTreeListNode()
         \         'size_with_commas=$(echo $size | sed -e :a -e "s/\(.*[0-9]\)\([0-9]\{3\}\)/\1,\2/;ta") && ' .
         \         'ls -ld ' . shellescape(treenode.path.str()) . ' | sed -e "s/ $size / $size_with_commas /"'
 
-        let metadata = systemlist(cmd)
+        let metadata = split(system(cmd),'\n')
         call nerdtree#echo(metadata[0])
     else
         call nerdtree#echo("No information available")
--- a/nerdtree_plugin/fs_menu.vim
+++ b/nerdtree_plugin/fs_menu.vim
@@ -211,9 +211,7 @@ function! NERDTreeListNode()
 
         let cmd = 'size=$(' . stat_cmd . shellescape(treenode.path.str()) . ') && ' .
         \         'size_with_commas=$(echo $size | sed -e :a -e "s/\(.*[0-9]\)\([0-9]\{3\}\)/\1,\2/;ta") && ' .
-        \         'ls -ld ' . shellescape(treenode.path.str()) . ' | sed -e "s/ $size / $size_with_commas /" && ' .
-        \         'unset size && ' .
-        \         'unset size_with_commas'
+        \         'ls -ld ' . shellescape(treenode.path.str()) . ' | sed -e "s/ $size / $size_with_commas /"'
 
         let metadata = systemlist(cmd)
         call nerdtree#echo(metadata[0])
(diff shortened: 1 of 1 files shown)


nerdtree_plugin/fs_menu.vim
```diff
@@ -203,10 +203,22 @@ endfunction
...
+            let stat_cmd = 'stat -c "%s" '
+        endif
+
+        let cmd = 'size=$(' . stat_cmd . shellescape(treenode.path.str()) . ') && ' .
+        \         'size_with_commas=$(echo $size | sed -e :a -e "s/\(.*[0-9]\)\([0-9]\{3\}\)/\1,\2/;ta") && ' .
+        \         'ls -ld ' . shellescape(treenode.path.str()) . ' | sed -e "s/ $size / $size_with_commas /" && ' .
+        \         'unset size && ' .
+        \         'unset size_with_commas'
+
+        let metadata = split(system(cmd),'\n')
```
lifecrisis: I would use `systemlist()` here.

nerdtree_plugin/fs_menu.vim
```diff
@@ -203,10 +203,22 @@ endfunction
 function! NERDTreeListNode()
     let treenode = g:NERDTreeFileNode.GetSelected()
     if treenode != {}
```
lifecrisis: I would use `empty()` here

nerdtree_plugin/fs_menu.vim
```diff
@@ -203,10 +203,22 @@ endfunction
...
+        \         'size_with_commas=$(echo $size | sed -e :a -e "s/\(.*[0-9]\)\([0-9]\{3\}\)/\1,\2/;ta") && ' .
+        \         'ls -ld ' . shellescape(treenode.path.str()) . ' | sed -e "s/ $size / $size_with_commas /" && ' .
+        \         'unset size && ' .
+        \         'unset size_with_commas'
+
+        let metadata = split(system(cmd),'\n')
         call nerdtree#echo(metadata[0])
     else
-        call nerdtree#echo("No information avaialable")
+        call nerdtree#echo("No information available")
```
lifecrisis: Good catch.. I never noticed this!

nerdtree_plugin/fs_menu.vim
```diff
@@ -203,10 +203,22 @@ endfunction
...
+        if has("osx")
+            let stat_cmd = 'stat -f "%z" '
+        else
+            let stat_cmd = 'stat -c "%s" '
+        endif
+
+        let cmd = 'size=$(' . stat_cmd . shellescape(treenode.path.str()) . ') && ' .
+        \         'size_with_commas=$(echo $size | sed -e :a -e "s/\(.*[0-9]\)\([0-9]\{3\}\)/\1,\2/;ta") && ' .
+        \         'ls -ld ' . shellescape(treenode.path.str()) . ' | sed -e "s/ $size / $size_with_commas /" && ' .
+        \         'unset size && ' .
```
lifecrisis: Do you need to unset `size` and `size_with_commas` here?  Note that this command is run within a subshell.

lifecrisis:
This looks somewhat brittle.  I wish I had more to offer in terms of suggestions.  Perhaps some alternative using a Vim builtin function like `getfsize()` is in order.

This doesn't change things for me since I rarely use this feature.  However, I would try to work this into a plugin if I could.

Your call.

nerdtree_plugin/fs_menu.vim
PhilRunninger: Nope. Those are leftover from my tweaking the command line in the shell.

PhilRunninger:
Yeah, I don't use this function much either. I'm just cleaning up an old PR that's gone stale.

PhilRunninger:
I think Martin considers this file and `exec_menuitem.vim` to be plugins already, not necessarily part of the core NERDTree, but bundled with it. It's kinda like `netrw` is with respect to Vim.

lifecrisis:
That's all okay! It's good to bundle these plugins, too.  It gives users examples to follow.

Can't test right now.  Will do soon, though.

PR #756, opened 2017-11-01
NERDTree keeps 'No Name' buffer after deleting current node 'md' #755 (closes #755)

### LINKED ISSUE #755: NERDTree keeps 'No Name' buffer after deleting current node 'md'
_To assist in resolving your issue, provide as much information as possible, in place of the ellipses (`…`) below._

---
**Environment:** _Describe your Vim/NERDTree setup._

>* Operating System: Archlinux, Linux Opensuse and Window 7
>* GVim version `:version`: 8.0 Included patch 1-69
>* Vim version `:version`: 8.0 Included patch 1-2
>* NERDTree version `git rev-parse --short HEAD`: f554c20
>* NERDTree settings applied in your vimrc, if any:
let NERDTreeWinPos = "left"
let NERDTreeWinSize = 35
let NERDTreeIgnore = ['\.job$', '^CVS$', '\.orig', '\.log$']
let g:NERDTreeDirArrows = 0
let g:NERDTreeStatusline = "%f"
let g:NERDTreeDirArrowExpandable = ''
let g:NERDTreeDirArrowCollapsible = ''

**Process:** _List the steps that will recreate the issue._

>0. Open several buffers from the tree
>1. Delete one file from the tree using md
   NERDTree ask for confirmation to delete the file and delete the associated buffer if open
>2. confirm delete file and buffer with 'y'

**Current Result:** _Describe what you you currently experience from this process._

> Buffer is actually deleted but it remains open as 'No Name'

To completely remove residue buffer one should call ':Bdelete'

/!\ I tested (re) this several time before submitting the issue and notice that sometime:
- residue buffer is replaced with newly open file
- If deleted buffer was the first one to be opened it will be removed completely without leaving any residue buffer
- **This is occurring only if the file to be deleted is the currently active buffer**

---
**Optional**

**Screenshot(s):**

image

**Possible Fix:** _(Have you poked around in the code?)_
The setting below doesn't affect the results.
```vim
"Automatically delete the buffer after deleting or renaming a file
if !exists("g:NERDTreeAutoDeleteBuffer")
  let g:NERDTreeAutoDeleteBuffer = 0
endif
```
There is one function to be reviewed
```vim
function! s:promptToDelBuffer(bufnum, msg)
```

I think the issue will be solved by changin the line below in **s:promptToDelBuffer**:
- from
```vim
exec "tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':enew! ' | endif"
```
- to
```vim
exec "tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':bnext! ' | endif"
```

mboughaba:
Go to the new buffer in buffer list instead of creating a new buffer.
If only one buffer is open, now we go back to nerdtree buffer.

### DIFF
--- a/nerdtree_plugin/fs_menu.vim
+++ b/nerdtree_plugin/fs_menu.vim
@@ -55,7 +55,14 @@ function! s:promptToDelBuffer(bufnum, msg)
         " Is not it better to close single tabs with this file only ?
         let s:originalTabNumber = tabpagenr()
         let s:originalWindowNumber = winnr()
-        exec "tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':bnext! ' | endif"
+        let s:listedBufferCount = len(getbufinfo({'buflisted':1}))
+        " Go to the next buffer in buffer list if at least one extra buffer is listed
+        " Otherwise open a new empty buffer
+        if s:listedBufferCount > 1
+            exec "tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':bnext! ' | endif"
+        else
+            exec "tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':enew! ' | endif"
+        endif
         exec "tabnext " . s:originalTabNumber
         exec s:originalWindowNumber . "wincmd w"
         " 3. We don't need a previous buffer anymore
--- a/nerdtree_plugin/fs_menu.vim
+++ b/nerdtree_plugin/fs_menu.vim
@@ -55,9 +55,17 @@ function! s:promptToDelBuffer(bufnum, msg)
         " Is not it better to close single tabs with this file only ?
         let s:originalTabNumber = tabpagenr()
         let s:originalWindowNumber = winnr()
-        let s:listedBufferCount = len(getbufinfo({'buflisted':1}))
         " Go to the next buffer in buffer list if at least one extra buffer is listed
         " Otherwise open a new empty buffer
+        if v:version >= 800
+            let s:listedBufferCount = len(getbufinfo({'buflisted':1}))
+        elseif v:version >= 702
+            let s:listedBufferCount = len(filter(range(1, bufnr('$')), 'buflisted(v:val)'))
+        else
+            " Ignore buffer count in this case to make sure we keep the old
+            " behavior
+            let s:listedBufferCount = 0
+        endif
         if s:listedBufferCount > 1
             exec "tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':bnext! ' | endif"
         else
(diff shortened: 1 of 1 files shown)


lifecrisis:
I'm leaning toward accepting this change. It avoids creating a new buffer, which I like.

@PhilRunninger, do you approve?

PhilRunninger:
This changes the current behavior when only one non-NERDTree buffer is open. It causes NERDTree to be the only window after `md`. This leads to window sizing issues that have caused a lot of confusion for other users. Maybe it would be better to leave the old behavior if deleting the last `listed` buffer. But I do like this change in all other situations.

mboughaba:
>This changes the current behavior when only one non-NERDTree buffer is open. It causes NERDTree to be the only window after md. This leads to window sizing issues that have caused a lot of confusion for other users. Maybe it would be better to leave the old behavior if deleting the last listed buffer. But I do like this change in all other situations.

Thanks guys! I agree.
I tried at least to put that in small words in my commit

> If only one buffer is open, now we go back to nerdtree buffer.

I'll work on that

nerdtree_plugin/fs_menu.vim
```diff
@@ -55,7 +55,14 @@ function! s:promptToDelBuffer(bufnum, msg)
         " Is not it better to close single tabs with this file only ?
         let s:originalTabNumber = tabpagenr()
         let s:originalWindowNumber = winnr()
-        exec "tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':enew! ' | endif"
+        let s:listedBufferCount = len(getbufinfo({'buflisted':1}))
```
lifecrisis: Note that `getbufinfo()` is not supported in Vim 7.4.

lifecrisis: I can't approve the change until this is fixed.

mboughaba: Thanks for the find. I wasn't aware that getbufinfo is not supported in 7.4. Will work on that.

mboughaba: Below works on vim 7.2 in my linux machine
```vim
:echo len(filter(range(1, bufnr('$')), 'buflisted(v:val)'))
```
I Need to find a clean way to do this

PhilRunninger: I'm fine with it either way - old syntax or both.

lifecrisis: @mboughaba, I'll eventually push for a fork that will break with Vim 7 and only support Vim 8. No point in doing it right now though, adoption is not wide enough to to warrant the move.

Believe me, if we break compatibility with 7 at this time, complaints will flow!

PR #759, opened 2017-11-11
BUGFIX: Repair broken "t" and "T" mappings

lifecrisis:
The title says it all. Pay particular attention to the commit message for ae1c95b.  That commit really fixes the bug for directories, and the following commit fixes the same issue for bookmarks.

It's funny that `open()` has a different signature for bookmark nodes than for file and directory nodes. This made the change more difficult to make general, but still doable.

### DIFF
--- a/lib/nerdtree/creator.vim
+++ b/lib/nerdtree/creator.vim
@@ -109,11 +109,6 @@ endfunction
 " FUNCTION: s:Creator._createNERDTree(path) {{{1
 function! s:Creator._createNERDTree(path, type)
     let b:NERDTree = g:NERDTree.New(a:path, a:type)
-    "TODO: This is kept for compatability only since many things use
-    "b:NERDTreeRoot instead of the new NERDTree.root
-    "Remove this one day
-    let b:NERDTreeRoot = b:NERDTree.root
-
     call b:NERDTree.root.open()
 endfunction
 
--- a/lib/nerdtree/creator.vim
+++ b/lib/nerdtree/creator.vim
@@ -47,33 +47,26 @@ endfunction
 " FUNCTION: s:Creator.createTabTree(a:name) {{{1
 " name: the name of a bookmark or a directory
 function! s:Creator.createTabTree(name)
-    let path = self._pathForString(a:name)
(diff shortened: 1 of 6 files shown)


lifecrisis:
Thanks, @PhilRunninger!!! As one last adjustment, do you think it would be a good idea to open tabs at the end of all open tabs, like with `:$tabnew`?  This has been suggested before in some issue or PR, and I shot it down... mainly because the user wanted an option. I almost think this should be the default... 

Either way, the tab can be moved after the fact to the desired location.  It's a minor change, but I would like to see it added.  Since the mappings were already broken, no one depended on them anyway...

PhilRunninger:
I have no opinion on this one, since I don't use tabs. Opening tabs at the end is consistent with other IDEs, so it's probably OK. Since is it new behavior, I would suggest putting a mention in the FAQ Wiki, explaining the change and the command to move tabs after they're created.

lifecrisis:
Got it.  I'll add it to the CHANGELOG as well.

lifecrisis:
The `CHANGELOG` has been updated and the change for new tabs has been made. So the issues are considered fixed.

I'll merge now and update the Wiki in a few minutes.

lib/nerdtree/creator.vim
```diff
@@ -110,11 +109,6 @@ endfunction
 " FUNCTION: s:Creator._createNERDTree(path) {{{1
 function! s:Creator._createNERDTree(path, type)
     let b:NERDTree = g:NERDTree.New(a:path, a:type)
-    "TODO: This is kept for compatability only since many things use
-    "b:NERDTreeRoot instead of the new NERDTree.root
-    "Remove this one day
-    let b:NERDTreeRoot = b:NERDTree.root
```
mboughaba: Hi @lifecrisis 
This change is causing Projectionist plugin to raise errors at startup. 
https://github.com/tpope/vim-projectionist/issues/89

I just want to highlight that but this is indeed something to be fixed in Projectionist

lifecrisis: Yup, we'll add that back. It's been a long time, so others should also be encouraged to update their code.

PR #777, opened 2017-12-01
Style Choice: Using confirm() when deleting a bookmark

lifecrisis:
This PR may warrant some debate.  I think we should use `confirm()` whenever possible...  This is a pilot of that idea.  I could see how someone would think a custom menu is better, though.

@PhilRunninger, let me know if you think these kinds of changes are a good idea!


PhilRunninger:
I like it. It feels like `confirm()` is specifically intended for this sort of thing. Not to say `getchar()` can't do it, but I like the way `confirm()` handles both the question and the answer.

Issue #779, opened 2017-12-10
A newly written file cannot be found with NERDTreeFind


lifecrisis:
To replicate:

1. Open Vim in home directory.
2. Open the NERDTree.
3. In the empty buffer to the right, run command `:w test.txt`.
4. Next, run `:NERDTreeFind`.
5. View the error... 

Looks like a refresh is needed before `:NERDTreeFind` is run.

lifecrisis:
Making a note here that several issues may be possible duplicates of this issue... no time to look into it right now.

juanibiapina:
I have the problem in my fork, but you can't just force a refresh because it might be very slow for some directory trees.

lifecrisis:
There are a few things that need to be dealt with here: 

- [ ] `findAndRevealPath()` needs to be updated to remove some duplication and use better style
- [ ] the documentation for `:NERDTreeFind` has to be modified to reflect the new behavior (directories as well as files can now be found and an argument can be passed)
- [ ] an efficient method needs to be developed to show the new file (i.e., something faster than a full refresh)

I'm doing some work on this right now...

lifecrisis:
For some reason this didn't get closed...

Closing now.

lifecrisis:
I had to reopen this issue because the fix that was added in #785 caused problems reported in #793.  A new fix will have to be devised and added later.

PhilRunninger:
This seems to be working now too.

juanibiapina:
When was the actual fix introduced?

PhilRunninger:
From what I can piece together, #785, merged on 12/22/17 ,fixed this issue, but introduced issue #793, which was reported on 1/2/18. PR #794 fixed that bug when it was merged on 1/5/18.

juanibiapina:
My understanding is that #794 removed the whole fix for this.

lifecrisis:
@PhilRunninger, the fix was removed.  I don't remember the reason, but I remember thinking it was a good idea at the time... so this is still broken, unfortunately.

juanibiapina:
This is how I fixed it on my fork: https://github.com/juanibiapina/vim-lighttree/blob/edd2da237f615441012679f47c371494966dafa1/autoload/lighttree.vim#L46

I just refresh the whole tree if the node cannot be found. For me that's a good trade off between speed and correctness, because the refresh only happens if the node cannot be found.

The code there has diverged considerably, but a similar fix should be possible here.

rstacruz:
[...]

PR #785, opened 2017-12-21
BUGFIX:  Allow ":NERDTreeFind" to reveal new files

lifecrisis:
Here, I'm addressing issue #779.

When a new file is written *after* the NERDTree for the current tab has been initialized, attempting to use `:NERDTreeFind` on that new file will fail with an error message. The reason for this is subtle, but it has to do with the following functions:

1. `NEDRTreeDirNode.reveal()`  (this is what really drives the `:NERDTreeFind` command)
2. `NERDTreeDirNode.open()`
3. `NERDTreeDirNode._initChildren()`

First, note that `reveal()` must `open()` all nodes along the path, starting with root, to "get to" the desired node.  Also note: `open()` calls `_initChildren()` **only if** the node's children *have not yet been initialized*.  This is for efficiency... but doesn't help much here because nodes will likely need to be opened anyway.  It only helps if one of the dirs along the path has *already* been opened and has, say 200+ children.

That last point is the key.  When the root is `open()`ed, it's children are *not* checked again.  However, if we create a new file several dirs deep (dirs that have not yet been opened), the `:NERDTreeFind` call will succeed because the dirs along the path are opened *after* the path was created.

In effect, the find command is searching the tree, not the file system... so some mechanism is needed to "freshly open" the nodes along the search path as we go.  We can be sure the path exists, we just need to make sure it's guaranteed to be in the tree when we get there!

The last commit on this branch fixes the issue in the most efficient way possible... but some may still find it undesirable... Read the comment in that commit when it shows up to see why it's most efficient... 

Oh, and feed back will be appreciated!

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -270,13 +270,13 @@ function! s:findAndRevealPath(path)
     endif
     
     try
-        let p = g:NERDTreePath.New(l:path)
+        let l:p = g:NERDTreePath.New(l:path)
     catch /^NERDTree.InvalidArgumentsError/
-        call nerdtree#echo("no file for the current buffer")
+        call nerdtree#echo('no file for the current buffer')
         return
     endtry
 
-    if p.isUnixHiddenPath()
+    if l:p.isUnixHiddenPath()
         let showhidden=g:NERDTreeShowHidden
         let g:NERDTreeShowHidden = 1
     endif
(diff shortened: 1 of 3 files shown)


autoload/nerdtree/ui_glue.vim
```diff
@@ -270,52 +270,50 @@ function! s:findAndRevealPath(path)
...
-            endif
+            call s:chRoot(g:NERDTreeDirNode.New(l:pathObj.getParent(), b:NERDTree))
         endif
     endif
-    call g:NERDTree.CursorToTreeWin()
-    let node = b:NERDTree.root.reveal(p)
+
+    let l:node = b:NERDTree.root.reveal(l:pathObj)
+
+    if empty(l:node)
```
PhilRunninger: Is this `if` block left over from testing? It seems out of place.

autoload/nerdtree/ui_glue.vim
```diff
@@ -270,52 +270,50 @@ function! s:findAndRevealPath(path)
...
+        echomsg 'l:node is totally empty...'
```
PhilRunninger: If this is to be kept, use the nerdtree function instead of `echomsg`.

lifecrisis:
@PhilRunninger... I hadn't pushed the fix yet!   Try it now!

lifecrisis:
@juanibiapina ... if you want to weigh in, let me know what you think.  I think this is as efficient as its gonna get.

lifecrisis:
Just a note... documentation still needs to be updated, and it looks like the function behaves improperly when the buffer doesn't have an associated path (try `:NERDTreeFind` in the quickfix window for example...).

PhilRunninger:
It looks like `:NERDTreeFind`, run from the Quickfix window, takes you to `cwd()`. What do you suggest it should be doing?

PhilRunninger:
Sorry. Wrong button.

lifecrisis:
Well, previously, `:NERDTreeFind` would tell you that no file existed for the current buffer (if it was a new buffer, say).  I think that should be the default behavior.  

If you specify a directory as an arg... then it should reveal the directory as expected.

That makes the most sense to me, anyway.

lifecrisis:
@PhilRunninger,

This pull request is O.K., but I'm not happy with the efficiency of my solution.  I don't really know a better way to do it... Here, I refresh the nodes along the path, which is better than refreshing the whole tree... however, it could be more efficient if I added just the new file to the tree (a single node).

I'm ambivalent, because the file we're finding could be in an unopened, but huge directory... which will be slow no matter what you do.  Thus, worrying about efficiency will be a waste of our time.

I've tried this on a directory of 350 nodes... it takes about 1.06s.  To me this is acceptable... if a developer is structuring their projects well, directories will be much smaller than this anyway.

Large dirs are just always going to be a problem.

What do you think?

PhilRunninger:
I agree. Large directories and vimscript itself both work against a perfect solution. I think we're OK to go with this.

lifecrisis:
Sounds good.  I'll review and merge in the next half-hour or so.

lifecrisis:
**NOTE:** This PR was reverted (but all style changes were kept) in PR #794.  The reasons for the change are given there.

Issue #189, opened 2012-08-22
NERDTreeFind does not mark hidden file


stabilobox:
For a hidden file the `:NERDTreeFind` will find the CWD but doesn't place the cursor on the file. Rather it marks the first directory entry.

E.g. open `~/.vimrc` and do `:NERDTreeFind`

A `:NERDTreeFind I` also does not do the trick.

mboughaba:
I get this issue too on GViM (Windows). It works fine on Linux.

lifecrisis:
Thanks for the report, guys. Just confirmed this bug. It will take some investigation to fix this properly.

lifecrisis:
@mboughaba and @stabilobox, you should both be able to update the NERDTree and have your problem fixed. Let me know if the issue is not corrected...

PR #786, opened 2017-12-22
BUGFIX:  Allow ":NERDTreeFind" to discover hidden files

lifecrisis:
Issue #189 originally pointed out a bug in which `:NERDTreeFind` could not always properly reveal a hidden file.  This PR adds a function that makes the behavior of `findAndRevealPath()` consistent in this respect.

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -277,11 +277,6 @@ function! s:findAndRevealPath(pathStr)
         return
     endtry
 
-    if l:pathObj.isUnixHiddenPath()
-        let l:showHidden = g:NERDTreeShowHidden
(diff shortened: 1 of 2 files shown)


lifecrisis:
@PhilRunninger, I think this is the last NERDTree code I'll write before the new year... it's a pretty good one to end on!

lifecrisis:
I'm going to go ahead and merge because I really want the functionality... this one has been bothering me for a while. I guess it's because I edit a lot of dot-files!

I've reviewed it several times so it should be good.

PhilRunninger:
D'oh! I forgot to review this one. I just ran a quick check of it, and it works properly on Mac.

lifecrisis:
Thanks, @PhilRunninger!!!

Issue #237, opened 2013-02-02
Question: Ability to always use existing window/buffer/tab when possible.


jordwalke:
Hello,

I've found the plugin Ctrl-p to pair very nicely with NERDTree. One thing that Ctrl-p does well, is that it has a mode where all open commands will find the existing file/buffer/tab whenever possible. This means that if you do a vertical split "s", it will only perform the vertical split if that file is not already present somewhere else. If it is present somewhere else, it instead will jump to that file. In other words, the type of open command "s/t" is just a fallback in case the file cannot be found globally. Is it possible to configure NERDTree to do the same?

PhilRunninger:
Closing for reasons cited in #791.

PR #791, opened 2017-12-30
Implements #237. Make split- and tab-open reuse windows.

PhilRunninger:
Instead of adding a new option or command to toggle the behavior, as
shown in #237, this commit makes the behavior the same as for the `o`
command for file nodes. If the file is already open in a window, it is
reused, instead of being opened in a new split or tab.

Since this is a change of behavior, splitting a window to duplicate
the buffer already in it will need to be done with the vim command
`:sp`, `:vsp`, or `:tabe %`.

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -110,7 +110,7 @@ endfunction
 "FUNCTION: s:activateBookmark() {{{1
 "handle the user activating a bookmark
 function! s:activateBookmark(bm)
-    call a:bm.activate(b:NERDTree, !a:bm.path.isDirectory ? {'where': 'p'} : {})
+    call a:bm.activate(b:NERDTree, !a:bm.path.isDirectory ? {'reuse': 'all', 'where': 'p'} : {})
 endfunction
 
 " FUNCTION: nerdtree#ui_glue#bookmarkNode(name) {{{1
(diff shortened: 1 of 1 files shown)


autoload/nerdtree/ui_glue.vim
```diff
@@ -509,7 +509,7 @@ endfunction
 
 " FUNCTION: s:openInNewTab(target) {{{1
 function! s:openInNewTab(target)
-    let l:opener = g:NERDTreeOpener.New(a:target.path, {'where': 't'})
+    let l:opener = g:NERDTreeOpener.New(a:target.path, {'reuse': 'all', 'where': 't'})
```
lifecrisis: @PhilRunninger,

I don't support merging this PR because I don't think the solution here is precise enough.

If I'm editing a file, and it's open in the current tab, pressing `t` on the node for that file in the tree should *force* the file to be opened in a new tab.  Reusing the open window with `o` is okay, but, when I say "Tab Page Please", I think I should get a new tab page!

I think the model for opening the file is okay the way it is, with `o` reusing the window, if it exists, and the other commands forcing specific opening behaviors.

lifecrisis: To add one more comment:

This is a case where I would prefer a new user preference... something like `g:NERDTreeAlwaysReuseWindows` as opposed to totally changing the behavior users have come to expect. That seems to be the only "precise" solution that avoids changing the way many users expect the NERDTree to behave.

PhilRunninger: Something @scrooloose mentioned a lot is the fact that NERDTree is already bloated, and he was reluctant to add options to satisfy everyone's whims. That's why I didn't make it configurable, but I see your point concerning the change of behavior. I guess I'd be a little put off if I had to change the way I open splits and tabs.

Do you think we should go ahead with the suggested new option, or just close this?

lifecrisis: I would prefer to just close.  I would only second adding the option if you were to argue that it was genuinely necessary.

Issue #147, opened 2012-03-05
Allow bookmarks to use environment variables


gburca:
I often work on 2 or more different versions of a project in parallel. Each terminal has an environment variable that defines the root of the project. It would be nice if I could define bookmarks like:

base $PROJ/frameworks/base

That bookmark would then point to a different location, depending on the terminal in which vim is running. Currently I have to maintain a separate bookmark for each project.

hiberabyss:
I run into the same problem when I switch from Ubuntu to Mac, please consider adding this useful feature. Thanks :)

PhilRunninger:
Closing this issue for the reasons mentioned in #792.

PR #792, opened 2018-01-02
Allow bookmarks to use environment variables

PhilRunninger:
Implements #147. Use the expand() function to interpret the environment variable contained in the bookmark's path.

### DIFF
--- a/lib/nerdtree/bookmark.vim
+++ b/lib/nerdtree/bookmark.vim
@@ -97,7 +97,7 @@ function! s:Bookmark.CacheBookmarks(silent)
 
                 let name = substitute(i, '^\(.\{-}\) .*$', '\1', '')
                 let path = substitute(i, '^.\{-} \(.*\)$', '\1', '')
-                let path = fnamemodify(path, ':p')
+                let path = fnamemodify(expand(path), ':p')
 
                 try
                     let bookmark = s:Bookmark.New(name, g:NERDTreePath.New(path))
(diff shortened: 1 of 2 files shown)


lib/nerdtree/bookmark.vim
```diff
@@ -97,7 +97,7 @@ function! s:Bookmark.CacheBookmarks(silent)
 
                 let name = substitute(i, '^\(.\{-}\) .*$', '\1', '')
                 let path = substitute(i, '^.\{-} \(.*\)$', '\1', '')
-                let path = fnamemodify(path, ':p')
+                let path = fnamemodify(expand(path), ':p')
```
lifecrisis: @PhilRunninger: My issue with this change is that `expand()` does more than just expand environment variables. For example, if a target path contains wildcard characters, the bookmark will never be interpreted as valid.

A simple case of this is a file named something like `*file*.txt` in your home directory with other files named `file1.txt` and `file2.txt`, for example.

Calling...

```viml
:echo expand('$HOME/*file*.txt')
```

... will give you more than you asked for.  It is unlikely, but it *is* a bug, IMHO.

We need to expand *only* environment variables.  I'm not sure it can be done without a considerable amount of hacking.  You also have the problem of directory names with dollar signs in them.  What if one happens to match an environment variable (again, not likely)?

I recommend to leave this PR out, unless these problems are securely addressed.

PhilRunninger:
No worries. I wasn't aware that `expand()` would do that. I'll close the PR and its issue.

PR #797, opened 2018-01-20
Fix broken keymaps (i/gi/s/gs/go) for bookmarks

suewonjp:
Currently, mappings of i,gi,s,gs,go are broken for bookmarks.

Documents or help mention nothing about these mappings on bookmarks, so I'm not sure these mappings are supposed to be provided.

However, there were code for them (with broken states) and currently, pressing those mappings will execute those broken code and give you error messages, which doesn't look good.

( I've tested my implementation and it looks good. )

If you reject this pull request, then at least removing those broken code is recommended so that users won't get confused

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -23,16 +23,16 @@ function! nerdtree#ui_glue#createDefaultBindings()
     call NERDTreeAddKeyMap({ 'key': g:NERDTreeMapOpenSplit, 'scope': "Node", 'callback': s."openHSplit" })
     call NERDTreeAddKeyMap({ 'key': g:NERDTreeMapOpenVSplit, 'scope': "Node", 'callback': s."openVSplit" })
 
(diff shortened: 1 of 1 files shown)


lifecrisis:
Thanks for the PR, @suewonjp.  This is a good observation.

@PhilRunninger, I just happened to log in at the right time and see this, thought I'd chime in...

This PR would need to include documentation updates to be acceptable (both for the quick help and the main help file).  However, I lean more toward just removing the mappings for the following two reasons:

1. They're undocumented, so few people know about them.
2. No one has complained about *not* having them until now.

I never even noticed that these mappings existed or were broken because I never used them or felt the need to use them. I don't ever see myself using them in the future either.

I recommend we close this PR.  @suewonjp, I would accept a PR that would delete the mappings to avoid the nasty (and unexpected) error messages.

suewonjp:
@lifecrisis 

I have no problem with closing it.

Well, just removing the broken code from the upstream master would be faster and simpler, I think.

Thank you.

PR #800, opened 2018-01-26
Make it possible to set the command to cd

grimpy:
Signed-off-by: Jo De Boeck <deboeck.jo@gmail.com>


lifecrisis:
I move to reject this pull request.  The functionality you're seeking can best be provided by a custom plugin.  Adding another global option is swimming against the direction the NERDTree needs to take.

For an example, see my plugin: [switch.vim](https://github.com/lifecrisis/vim-config/blob/master/nerdtree_plugin/switch.vim).

In my plugin...
* `ct` will change the tab-local working directory (note that it works in Vim, not just NeoVim... read the code to find out how)
* `cw` will change the window-local working directory
* `cd` will change the global directory

Feel free to use it and share it... the current version is in the public domain.  Tab-local working directories don't really make sense. The `:windo` command already makes this possible.

@PhilRunninger, let me know if you disagree.  Otherwise, close at your discretion.

PR #801, opened 2018-01-27
Fix focussing previous buffer when closing NERDTree

mrubli:
Previously closing NERDTree while two windows were showing the same buffer would
focus the first window, which was not necessarily the previously active one.

Instead of obtaining the buffer ID of the previous buffer and mapping that to
the window ID (which is a 1:n mapping) we obtain the unique window ID and focus
the right window after closing NERDTree.

win_getid() and win_gotoid() are available from VIM 7.4.1557, so this patch is
conditional on that version.

### DIFF
--- a/lib/nerdtree/nerdtree.vim
+++ b/lib/nerdtree/nerdtree.vim
@@ -38,17 +38,26 @@ function! s:NERDTree.Close()
     endif
 
     if winnr("$") != 1
+        " Use the window ID to identify the currently active window or fall
+        " back on the buffer ID if win_getid/win_gotoid are not available, in
+        " which case we'll focus an arbitrary window showing the buffer.
+        let l:useWinId = exists('*win_getid') && exists('*win_gotoid')
+
         if winnr() == s:NERDTree.GetWinNum()
             call nerdtree#exec("wincmd p")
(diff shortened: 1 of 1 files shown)


lifecrisis:
This is a good observation.  The style is not what I would like.

Some notes:
* on line 50 you use `win_getid()` without a version check
* you should use a funcref with `exists()` in this PR... that's the nicest way to see if you have the feature
* you should wrap your commit message at 68 characters

This could be improved.  I'll fix it myself in the future if you're not interested in working on it further.

mrubli:
Thanks for the comments and good catch on the missing version check; I'll rework the PR later this week to follow your suggestions. I rarely write Vimscript, so when it comes to the nicest way of doing things I have some catching up to do. :-)

mrubli:
@lifecrisis I pushed an updated patch that hopefully addresses your notes. Please take a look.

PhilRunninger:
@mrubli , thanks for the contribution. I'm curious why your first commit doesn't show in the pull request anymore.

mrubli:
@PhilRunninger Given the issues it had I just amended the initial commit. I also had the wrong author info in there because I happened to commit from my work computer. It's much cleaner this way.

PR #809, opened 2018-02-26
Fix error if not in UTF-8

HadesD:
My PC not run in UTF-8, so it shown error and can not use NerdTree.

### DIFF
--- a/plugin/NERD_tree.vim
+++ b/plugin/NERD_tree.vim
@@ -69,7 +69,7 @@ call s:initVariable("g:NERDTreeShowHidden", 0)
 call s:initVariable("g:NERDTreeShowLineNumbers", 0)
 call s:initVariable("g:NERDTreeSortDirs", 1)
 
-if !nerdtree#runningWindows() && !nerdtree#runningCygwin()
+if !nerdtree#runningWindows() && !nerdtree#runningCygwin() && &encoding == "utf-8"
     call s:initVariable("g:NERDTreeDirArrowExpandable", "▸")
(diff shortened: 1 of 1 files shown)


PhilRunninger:
@HadesD Could you provide some more details about the situation this pull request aims to correct? What OS are you running? Are you using ssh or something similar? Are you working in a language other than English? What is your `&encoding` set to? Any other information that might be useful?

I also wonder if `&& &encoding == "utf-8"` is too restrictive. Would `utf-16` or others still enable the triangle characters to work?

HadesD:
Hi @PhilRunninger 
My problem is, my current OS run in EUC-JP, and alots of projects are EUC-JP, so it does not support the UTF-likes character.

So I think that we should add an options like 
```
let g:default_character = {...}
```

PhilRunninger:
I'm not sure I want to merge this one just yet. I'd like to get a review from @lifecrisis before proceeding.

In the meantime, you can manually set those values in your `.vimrc` file:
```
let NERDTreeDirArrowExpandable = '+'
let NERDTreeDirArrowCollapsible = '~'
```
Use the characters provided by your encoding and font to set these arrows as you like.

lifecrisis:
Having `'encoding'` set to "utf-8" does not guarantee that the underlying system actually supports the necessary Unicode code points (though it may support a broad range of Unicode symbols).  There is simply no good way to solve this problem, so a default is selected that works and looks nice in most places.  It is up to the user to configure their settings to their liking beyond that.  I see no cogent argument for why we should change.

We appreciate the idea, but closing this pull request is reasonable.

PR #816, opened 2018-03-13
Add new variable g:NERDTreeRemoveFileCmd

kutsan:
`let g:NERDTreeRemoveFileCmd = 'trash '` and everyone is happy.

### DIFF
--- a/doc/NERDTree.txt
+++ b/doc/NERDTree.txt
@@ -694,6 +694,12 @@ NERD tree. These options should be set in your vimrc.
 |'NERDTreeCreatePrefix'|        Specify a prefix to be used when creating the
                                 NERDTree window.
 
+|'NERDTreeRemoveFileCmd'|       Specify a custom shell command that will used
+                                when deleting files with delete action. Note
+                                that, it should be one space character at the
+                                end of the command and it applies only for files.
+                                See |g:NERDTreeRemoveDirCmd| for directories.
+
 ------------------------------------------------------------------------------
 3.2. Customisation details                             *NERDTreeOptionDetails*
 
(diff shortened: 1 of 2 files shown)


PhilRunninger:
@kutsan What problem is this code solving? Does vim's delete() function not do what it's supposed to do?

kutsan:
@PhilRunninger vim's `delete()` function deletes given item like built-in `rm` command does, so it's not possible to recover since it's actually gone after invoking it. There are a lot of alternatives that adds functionality to move given item to trash. I use [trash-cli](https://github.com/sindresorhus/trash-cli) for that. If you happen to delete a file with `md` by mistake, you can still recover from trash with `let g:NERDTreeRemoveFileCmd = 'trash '`.

PhilRunninger:
That's what I suspected. Add some text to the help file, and I'll merge this.

kutsan:
Thanks! Does it look good now?

Also there isn't any documentation for this variable:

https://github.com/scrooloose/nerdtree/blob/ed446e5cbe0733a8f98befc88d33e42edebb67d2/plugin/NERD_tree.vim#L108

kutsan:
And by the way, I really don't like having a need for a space character at the end of the command because we can do that programmatically but original author chose to not to and it's like that for 10 years. Now sadly, we can't go back for the sake of backwards compatibility.

PhilRunninger:
Thanks for the heads up on the missing doc. I'll add a sentence after I merge.

kutsan:
Also, missing:

https://github.com/scrooloose/nerdtree/blob/ed446e5cbe0733a8f98befc88d33e42edebb67d2/plugin/NERD_tree.vim#L109-L110

PhilRunninger:
I'll add those too, soon. I'll also remove the need for the trailing space.

PR #829, opened 2018-04-22
Display a path with multi-byte characters correctly when it is truncated

bravestarr:
When a path is rendered in the Bookmarks area or as the header line of the
tree, it is truncated if there is not enough space for it.  But if a path
contains multi-byte characters, it should be truncated by characters, not
bytes, otherwise the path may be truncated between the bytes of a
multi-byte character.  To deal with multi-byte characters, use
strdisplaywidth() instead of len() to get the number of display cells, and
use strcharpart() instead of strpart() to truncate a path.


bravestarr:
Reopen the commit

PR #830, opened 2018-04-22
Display a path with multi-byte characters correctly when it is truncated

bravestarr:
When a path is rendered in the Bookmarks area or as the header line of the
tree, it is truncated if there is not enough space for it.  But if a path
contains multi-byte characters, it should be truncated by characters, not
bytes, otherwise the path may be truncated between the bytes of a
multi-byte character.  To deal with multi-byte characters, use
strdisplaywidth() instead of len() to get the number of display cells, and
use strcharpart() instead of strpart() to truncate a path.


PhilRunninger:
@bravestarr , I'm not sure how to test this. I tried creating a file with a mix of characters from both ends of the Latin set, one with all non-ASCII characters, and even one that was all emojis. Can you provide some screenshots or instructions for recreating?

Thanks,

bravestarr:
I have reproduced the problem by making a directory as a bookmark of which the path contains Chinese characters.  The full path is E:\for-nerdtree\我的书签.  I have no idea if you have any Chinese fonts installed, if not, the four Chinese characters are just as in the following screenshots.  Each Chinese character is encoded as three bytes (UTF-8), so the four characters are \<E6 88 91\>, \<E7 9A 84\>, \<E4 B9 A6\>, \<E7 AD BE\>.

When no truncation is needed, all goes well:
all-is-good

Change the width of the window for nerdtree, and double-click the bookmark to force the window to be redrawn.  When the leading byte of the first character is truncated, the path is shown as:
bookmark-1
Note that the remaining two bytes of the first character are rendered as their hex representation.

Make the window even narrower, when the leading two bytes are truncated:
bookmark-2

The head line of the tree has the similar problem:
tree-1
tree-2

I also tried a path with Greek letters, the problem remained too.  I think the problem should occur when a path contains any multi-byte characters.  @PhilRunninger , You said you had tried non-ASCII characters, if the problem didn't appear, I guess you may not change the width of the window to the degree that only the leading bytes for a character are truncated, or you may not double-click the bookmark to force the window to be redrawn.

PR #833, opened 2018-04-25
Fixed error 'unknown function strcharpart' for older versions of Vim

hav4ik:
Previous commit ([`d3a7cd2`][prev]) resulted in a immediate crash when NERDTree is started, as shown in the screenshot below:

screenshot

This is due to the function `strcharpart()`, which has been added to vim in commit [vim/vim@`58de0e2`][vim] on April 14 2016 which is equivalent to the patch level 7.4.1730. One of the most popular operating systems, Ubuntu 16.04 LTS, is still using vim 7.4.1689, which doesn't have this function.

So I added backward-compatibility, by checking `if exists("*strcharpart")`

[prev]:https://github.com/scrooloose/nerdtree/commit/d3a7cd20aed896e62ca4fbcd44b8ebdf842d1adf
[vim]:https://github.com/vim/vim/commit/58de0e2dcc1f2d251b74892a06d71a14973f3187

### DIFF
--- a/lib/nerdtree/bookmark.vim
+++ b/lib/nerdtree/bookmark.vim
@@ -287,10 +287,14 @@ function! s:Bookmark.str()
 
     let pathStr = self.path.str({'format': 'UI'})
     if strdisplaywidth(pathStr) > pathStrMaxLen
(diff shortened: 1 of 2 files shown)


PhilRunninger:
@hav4ik Thanks for fixing that oversight.

PR #834, opened 2018-04-26
Replace strcharpart() with substitute() for backward compatibility

bravestarr:
Pull request #833 has supplied a fix to workaround the backward compatibility problem in #830.  But the multi-byte path problem will appear again in the older version of vim before the function strcharpart was introduced.  We can use substitute to achieve the same result as when strcharpart is used.

### DIFF
--- a/lib/nerdtree/bookmark.vim
+++ b/lib/nerdtree/bookmark.vim
@@ -288,7 +288,7 @@ function! s:Bookmark.str()
     let pathStr = self.path.str({'format': 'UI'})
     if strdisplaywidth(pathStr) > pathStrMaxLen
         while strdisplaywidth(pathStr) > pathStrMaxLen && strchars(pathStr) > 0
-            let pathStr = strcharpart(pathStr, 1)
+            let pathStr = substitute(pathStr, '.\{1}', '', '')
         endwhile
         let pathStr = '<' . pathStr
     endif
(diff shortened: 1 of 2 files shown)


lib/nerdtree/bookmark.vim
```diff
@@ -288,7 +288,7 @@ function! s:Bookmark.str()
     let pathStr = self.path.str({'format': 'UI'})
     if strdisplaywidth(pathStr) > pathStrMaxLen
         while strdisplaywidth(pathStr) > pathStrMaxLen && strchars(pathStr) > 0
-            let pathStr = strcharpart(pathStr, 1)
+            let pathStr = substitute(pathStr, '.\{1}', '', '')
```
PhilRunninger: I think the regex `'^.'` would be safer to use here. That guarantees it finds the first character.

lib/nerdtree/path.vim
```diff
@@ -721,7 +721,7 @@ function! s:Path.str(...)
         let limit = options['truncateTo']
         if strdisplaywidth(toReturn) > limit-1
             while strdisplaywidth(toReturn) > limit-1 && strchars(toReturn) > 0
-                let toReturn = strcharpart(toReturn, 1)
+                let toReturn = substitute(toReturn, '.\{1}', '', '')
```
PhilRunninger: Same regex comment here.

bravestarr:
I have commit this change to the branch used in #830, but the branch didn't sync to the master branch, so I did't mention the removal of `if exists("*strcharpart")`.  Now I have resolved the conflicts and used the regex you advised.

Issue #244, opened 2013-03-12
Always opening NERDTree in current window and reuse instance


marlun:
I read http://vimcasts.org/blog/2013/01/oil-and-vinegar-split-windows-and-project-drawer/ and as Drew says you can add `let NERDTreeHijackNetrw=1` and NERDTree will open up in the current window if I do `:e .`. However it seems to be a new instance of NERDTree every time it's opened because it does not remember which folders I had open before like it does if I run `:NERDTreeToggle`. Is it possible to have both? Have NERDTree open in the current window and remember where it was (reuse the same instance)?

knubie:
Every time NERDTree creates an instance via the NERDTreeHijackNetrw function is assigns it a unique buffer name i.e. `NERD_tree_1`, `NERD_tree_2`, etc. Why it's designed this way is not entirely clear to me, but as it turns out you can easily switch to the first instance by typing `:buffer NERD_tree_1` without ever having to create a new one.

So instead of editing the NERDTree source, I just created a mapping that looks like this:
```
nnoremap <leader>e :buffer NERD_tree_1<CR>
```
Hope this helps.

pass86:
I like this feature too!

scrooloose:
Hi, I have added a simple proof of concept for this in this branch: https://github.com/scrooloose/nerdtree/tree/reuse-win-tree

Try it out and let me know how you get on :)

scrooloose:
@PhilRunninger any thoughts on this? It seems like The Right Way (TM) to go - at least in theory.

PhilRunninger:
I like the approach. I'm getting an error though. I think it's related to buffer 1 no longer being in my list as you loop through the list.

After entering `:e ..` for the second (maybe third) time...
![image](https://cloud.githubusercontent.com/assets/5598066/19409949/ffe3542c-92ad-11e6-977a-5c53d501c856.png)

scrooloose:
Hmm, I cant reproduce this :-/

Can you give any more info about it?

PhilRunninger:
1. `:e ..` opens a NERDTree
1. Expand a folder, and open a file.
1. `:e ..` again should open the same NERDTree buffer, but I get the error shown above.

It happens on the `let nt = getbufvar(i, "NERDTree")` statement when nt exists from the previous loop iteration, and is an incompatible type. If it's already an empty string, it cannot be set to a dictionary, and vice versa.

I added a new commit to fix it, but I don't know if I did it the correct way.

[...]

PR #839, opened 2018-05-15
Add an option to open NERDTree in current window

zefei:
In a multi-window workflow, it's often desirable to make NERDTree to
work in the current window instead of its side window. See issue #244
for brief discussion.

Currently it's possible to use `:e <dir>` to achieve this for
`:NERDTree` command; but there are no workarounds for other functions
like opening bookmark, toggling nerdtree, finding current file, etc.

This commit adds an option to render NERDTree in current window instead
of side window, and modifies all relevant commands to respect the
option.

### DIFF
--- a/autoload/nerdtree.vim
+++ b/autoload/nerdtree.vim
@@ -10,40 +10,16 @@ endfunction
 " SECTION: General Functions {{{1
 "============================================================
 
-"FUNCTION: nerdtree#checkForBrowse(dir) {{{2
-"inits a window tree in the current buffer if appropriate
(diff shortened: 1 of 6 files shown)


lifecrisis:
This is not a pull request I want to merge.

We're trying not to add any more global options.  So, I suggest that you add this functionality by extending the NERDTree with a plugin for your own private use.  This PR adds a lot of code, which is no good for an already bloated project.

I don't mean to be harsh, but this is not the direction in which I want the NERDTree to go.  Feel free to submit PRs in the future.  Bug-fixes are especially welcome at this time.

Issue #250, opened 2013-04-22
<C-O> doesn't work in NERDTree


qstrahl:
# Procedure
1. I am in file `/var/www/example`
2. I jump to directory `/var/www/derpherper/`
3. I hit `<C-O>` to go back to `/var/www/example`
# Erroneous Result

The cursor is placed on the last line in NERDTree
# Expected Result

Vim jumps back to `/var/www/example`

juanibiapina:
There are many bugs involving ctrl+o and nerd tree, it becames unusable if you're using `:e .`

lifecrisis:
I'm unable to replicate this problem right now.  If either of you could provide more specific examples, I'd be willing to try and confirm.

juanibiapina:
Steps:

- clear jump list
- create two files called `1` and `2`
- `:e 1`
- `:e 2`
- `:e .` (with hijack, opens nerd tree)
- ctrl+o (goes to end of nerd tree, annoying but fine)
- ctrl+o (goes to file 1, wrong, end of jumps)
- ctrl+i (goes to file 2, correct)

After that you can play around with ctrl+o and ctrl+i and even though the jump list seems to be correct, it jumps to weird places.

I just added this line from the documentation which seems to improve the behavior for my example tests:
```
let NERDTreeCreatePrefix='silent keepalt keepjumps'
```

Although I still get jumps back to nerd tree and inside of it. I'm not sure I want this behavior. I'll live with my fix for a while and figure out later if I want nerdtree to mess with the jumplist at all.

lifecrisis:
Very good work, @juanibiapina.  This bug is confirmed.

lifecrisis:
I discovered another instance of this problem just now:

1. Position the cursor on a directory node that is no the root.
2. Press `P`.
3. Press `R` to refresh the tree.
4. Press `<C-O>` and notice that the cursor jumps to the end of the NERDTree window.

There is clearly a problem with the way that the NERDTree does not preserve jumps when printing the tree to the tree window.

juanibiapina:
The worst case is when ctrl+o does not work at all, even though jumplist is full of stuff. I can't reproduce it in isolation, although it happens to me everyday.

juanibiapina:
[...]

PR #844, opened 2018-05-26
Use ":clearjumps" in new NERDTree windows

lifecrisis:
See commit message.


PhilRunninger:
I'm not sure about this. Looking at @juanibiapina's comments in #250, there may be cases where jumping out of NERDTree with `<C-O>` is desirable, particularly when you use `:e .`. If I've misunderstood either of these scenarios, let me know.

PR #851, opened 2018-06-11
Decrease startup-time by avoiding linear-time iteration over maps

mnussbaum:
This commit makes adding, removing and finding key maps an O(1) operation instead of O(n), where n is the number of pre-existing maps.

In my testing, averaged over 100 iterations, this reduces the time spent initializing NERDTree at Vim startup from ~73ms to ~9.7ms. That's with only the default included key maps.

I chose to add an internal `s:KeyMap.all` function to initialize and return the key map dictionary so that the public `s:KeyMap.All` function's signature stays the same, returning a sorted list of key maps.

Let me know if there's any revisions you'd like to see!

### DIFF
--- a/lib/nerdtree/key_map.vim
+++ b/lib/nerdtree/key_map.vim
@@ -3,16 +3,21 @@
 let s:KeyMap = {}
 let g:NERDTreeKeyMap = s:KeyMap
 
-"FUNCTION: KeyMap.All() {{{1
-function! s:KeyMap.All()
-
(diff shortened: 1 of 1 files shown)


mnussbaum:
I think this fixes the majority of the startup time issue https://github.com/scrooloose/nerdtree/issues/276

mnussbaum:
Thanks! Just pushed 657be6b  to update the private function to add the leading underscore

PR #852, opened 2018-06-12
Bugfix - ensure keymaps dictionary exists before using it

mnussbaum:
I realized I might have introduced a bug in https://github.com/scrooloose/nerdtree/pull/851, I haven't been able to trigger it but I wanted to be safe rather then sorry.

The `KeyMap.Add` and `KeyMaps.Remove` functions can't use `KeyMap._all` to fetch the key map dictionary since they need to modify it, and it appears the dictionary returned by the `KeyMap._all` function is a copy, and modifying the copy doesn't modify the underlying `s:keyMaps` variable. However, using the `s:keyMaps` variable directly is error prone since the variable will be undefined until the very first invocation of `KeyMap._all`.

This PR invokes and discards the result of `KeyMap._all` prior to adding or removing any key maps from `s:keyMaps`directly, just to ensure `s:keyMaps` is defined.

### DIFF
--- a/lib/nerdtree/key_map.vim
+++ b/lib/nerdtree/key_map.vim
@@ -2,19 +2,11 @@
 "============================================================
 let s:KeyMap = {}
 let g:NERDTreeKeyMap = s:KeyMap
-
-"FUNCTION: KeyMap._all() {{{1
-function! s:KeyMap._all()
-    if !exists("s:keyMaps")
-        let s:keyMaps = {}
-    endif
-
-    return s:keyMaps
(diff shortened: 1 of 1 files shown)


PhilRunninger:
Then we wouldn't need the `s:KeyMap._all()` function either. Just use the `s:keyMaps` dictionary instead.

mnussbaum:
Good call, I've amended the commit to just initialize the `s:keyMaps` dictionary at the top of the file and totally removed the `KeyMap._all` function

PR #853, opened 2018-06-15
Better OSX detection

bubba-h57:
Neovim does not support an *osx* feature to test for. This PR implements a more generic test for the OSX operating system in order to ensure the correct `stat` command is used.

### DIFF
--- a/nerdtree_plugin/fs_menu.vim
+++ b/nerdtree_plugin/fs_menu.vim
@@ -228,7 +228,7 @@ function! NERDTreeListNode()
         if has("unix")
             let s:uname = system("uname")
             let stat_cmd = 'stat -c "%s" ' 
-            if s:uname == "Darwin\n"
+            if s:uname =~? "Darwin"
                 let stat_cmd = 'stat -f "%z" '
             endif
         endif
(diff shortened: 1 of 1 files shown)


nerdtree_plugin/fs_menu.vim
```diff
@@ -225,10 +225,12 @@ endfunction
...
     let treenode = g:NERDTreeFileNode.GetSelected()
     if !empty(treenode)
-        if has("osx")
-            let stat_cmd = 'stat -f "%z" '
-        else
-            let stat_cmd = 'stat -c "%s" '
+        if has("unix")
+            let s:uname = system("uname")
+            let stat_cmd = 'stat -c "%s" ' 
+            if s:uname == "Darwin\n"
```
PhilRunninger: Change this operator to `=~?` and remove the `\n`. The case-insensitive regex will help future-proof the if statement. See `:h expr-=~?`.

nerdtree_plugin/fs_menu.vim
```diff
@@ -225,10 +225,12 @@ endfunction
 function! NERDTreeListNode()
     let treenode = g:NERDTreeFileNode.GetSelected()
     if !empty(treenode)
-        if has("osx")
-            let stat_cmd = 'stat -f "%z" '
-        else
-            let stat_cmd = 'stat -c "%s" '
+        if has("unix")
```
PhilRunninger: Taking a wider look at this file, I noticed that this `if has("unix")` statement has already been done on line 41, so we don't need it again here.

PhilRunninger:
Thanks, @bubba-h57, for the improvement. Good catch!

Issue #842, opened 2018-05-21
should compare with type instead of its value


linuxholic:
https://github.com/scrooloose/nerdtree/blob/6188c5ea5f6fd44e24768ccb889aaa739808d9f3/autoload/nerdtree.vim#L76

PhilRunninger:
Thanks for the information. There's another issue, #277, that's probably closely related to this. I'll be taking a closer look at these soon.

PhilRunninger:
@linuxholic , Take a look at your scenario, and see if #856 has fixed it.

Issue #277, opened 2013-07-29
NERDTreeSortOrder doesn't seem to be respected after the first element in the list.


tankorsmash:
If I alter the sort order to something like `['\.cpp$', '\/$']` to have the `cpp` files listed before the directories, only the `cpp` files are listed first, and then the other files, along with the directories, are listed in alphabetical order. 

http://i.imgur.com/kJ0fLX3.png

At first I thought it was a typo or something similar, but if I re-order the list, so that the directories are first, and the `cpp` files are second, the directories are correctly placed first, but the `cpp` files are mixed in with the rest of the files in the directory.

http://i.imgur.com/tuQiUHY.png

PhilRunninger:
Confirmed.

PhilRunninger:
This should be fixed with #856. @tankorsmash , take another look at this, and reopen this issue if necessary.

tankorsmash:
Thanks @PhilRunninger!

PR #856, opened 2018-07-02
Fix issues with sorting of nodes

PhilRunninger:
This  PR fixes two issues: 
1. #842  (See `autoload/nerdtree.vim`)
> The sort algorithm was incorrectly comparing the value in the sort key to the type of the integer 0. Instead it should have been checking the type of the sort key's value to see it is numeric.

1. #277  (See `lib/nerdtree/path.vim`, `lib/nerdtree/tree_dir_node.vim`, and `plugin/NERD_tree.vim`)
> This issue reported that NERDTree was respecting only the first item in the sort order. It was probably more accurate to say that **changing the NERDTreeSortOrder in the current session, then refreshing NERDTree** didn't produce correctly sorted nodes. When doing this, NERDTree was using the cached value for the nodes' sortKey, which could be different than it needed to be for the new SortOrder. To fix this, I changed the code to always calculate the sortKey, instead of relying on a cached value. This adds some processing time, which may or may not be deemed acceptable.

### DIFF
--- a/autoload/nerdtree.vim
+++ b/autoload/nerdtree.vim
@@ -61,7 +61,6 @@ endfunction
 function! nerdtree#compareNodesBySortKey(n1, n2)
     let sortKey1 = a:n1.path.getSortKey()
     let sortKey2 = a:n2.path.getSortKey()
-
     let i = 0
     while i < min([len(sortKey1), len(sortKey2)])
         " Compare chunks upto common length.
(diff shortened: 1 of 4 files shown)


PhilRunninger:
@lifecrisis, Take a look at this now. I believe this will be sufficiently fast enough. The sort key will be recalculated only if it's uninitialized or when the g:NERDTreeSortOrder changes. Using your numbers above, your 100 root refreshes should take only `1 * 70.61 / 100  +  99 * 55.09 / 100` or `55.25` seconds.

lifecrisis:
This works!  I'm no longer seeing the big slow-down I noticed before.

I did only a perfunctory review of the changes to functionality, so I'm mainly just trusting you on that front.  I don't make use of this feature myself.  Thus, I haven't tested it much "in the field."

Unless you have other qualms about this change, feel free to merge!

PR #866, opened 2018-08-02
Default to placing cursor on root when closing bookmark table

lifecrisis:
Don't know why this bothered me so much... but it *needed* to be fixed.

### DIFF
--- a/lib/nerdtree/ui.vim
+++ b/lib/nerdtree/ui.vim
@@ -475,15 +475,23 @@ function! s:UI.toggleIgnoreFilter()
 endfunction
 
 " FUNCTION: s:UI.toggleShowBookmarks() {{{1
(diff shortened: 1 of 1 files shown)


PhilRunninger:
@lifecrisis , I checked on both Mac and Windows, and I don't see any difference between your branch's behavior and that of master's. What steps should I use to recreate the original issue?

lifecrisis:
1. Toggle on bookmark table.  (it helps if you have several bookmarks)
2. Navigate to the last entry in the table.
3. Toggle off the bookmark table.
4. You'll see that the cursor just goes wherever instead of being placed on the tree root.

If the cursor was already on a tree node, we want it to remain on that tree node (which it does with this fix).  However, if the cursor was somewhere above the tree, we now just put it on the tree root.

PR #867, opened 2018-08-04
Reword renamed-buffer prompt to be more clear

aflock:
When renaming a file that is currently open in a buffer, NerdTree helpfully asks us:
```
Node renamed. The old file is open in buffer <buf>. Replace this buffer with *a* new file? (yN)
```
(emphasis mine)

This is slightly confusing as replacing the buffer with **a** new file isn't what is meant, instead we can be more clear just by saying 'the new file', the one we just renamed to. 
```
Node renamed. The old file is open in buffer <buf>. Replace this buffer with the new file? (yN)
```


PhilRunninger:
Good catch. Thanks for the contribution.

Issue #680, opened 2017-04-04
Cannot open file or directory whose name include [ or ]


faaaar:
MacOS 10.12.3
Vim 8.0
scrooloose/nerdtree installed via vundle

lifecrisis:
Confirmed this bug on my installations.

PhilRunninger:
I believe this is only a problem when the file contains both brackets, with [ followed by ]. I'm not even seeing the file in the NERDTree. #391 may be the same issue.

lifecrisis:
Hey, @PhilRunninger! I've isolated the bug to the `s:UI._stripMarkup()` in `ui.vim`. It was pretty clear that files like this must be treated like flags, so I looked there. I'm fairly confident that [line 389](https://github.com/scrooloose/nerdtree/blob/master/lib/nerdtree/ui.vim#L389) in `ui.vim` is the source of our problem.

I'm not a regular expression wizard yet. Maybe you can take a look.

PhilRunninger:
It looks like [line 389](https://github.com/scrooloose/nerdtree/blob/master/lib/nerdtree/ui.vim#L389) is used to remove the flag that git-nerdtree-plugin puts on, but if **`foo[bar].txt`** is not a git-controlled file, it removes the wrong thing. If we force **`substitute()`** to look at the beginning of the word only, then we run into problems with **`[foo]bar.txt`**. My initial thought was to use **`filereadable()`**, and skip the substitute if it's not needed.

lifecrisis:
That actually sounds like a good decision for now. It seems more like a temporary fix, because other flags can exist that cause `filereadable()` to fail. We might need a more "heavyweight" solution for a long-term fix.

xcambar:
I've been hit with this bug and even though I agree that the problem requires a significant change to be rock-solid, but I've found a minimal change that works under the (weak) assumption that flags **generally** use special characters.

Here's the patch, for those interested:

```patch
diff --git a/lib/nerdtree/ui.vim b/lib/nerdtree/ui.vim
index 196d745..07a87e6 100644
--- a/lib/nerdtree/ui.vim
+++ b/lib/nerdtree/ui.vim
@@ -381,7 +381,7 @@ function! s:UI._stripMarkup(line)
     let line = substitute (line, '*\ze\($\| \)', "","")
 
     " strip off any generic flags
-    let line = substitute (line, '\[[^]]*\]', "","")
+    let line = substitute (line, '\[\W*\]', "","")
 
[...]

Issue #555, opened 2016-02-26
Directories with `{}` characters display strangely


jsdevel:
![nerdtree](https://cloud.githubusercontent.com/assets/2605980/13368349/d1839500-dca6-11e5-8a04-eca105a23b9f.png)

lifecrisis:
Confirmed bug. Has to do with highlight regular expression confusing the directory name with a bookmark flag.

PhilRunninger:
@jsdevel , would you be willing to test the pull request #868 I've put together for this issue? Details are in the PR. Thanks.

jsdevel:
@PhilRunninger sure!  may take me a couple days though

jsdevel:
@PhilRunninger I checked out that branch.  Doesn't seem to have fixed it.

![not-working](https://user-images.githubusercontent.com/2605980/44276487-7580b580-a1fc-11e8-98bd-0139bae2ad83.png)

jsdevel:
Why is it showing . and .. directories?

PhilRunninger:
I agree; the . and .. directories shouldn't be displayed. I couldn't find a way to reproduce this though. What environment are you in? OS? Shell? Can you share the complete directory listing with as much detail as possible as it appears in the shell? Does this happen in all NERTree directories, or just {foo}?

jsdevel:
Fedora Linux, Terminator, happens any time Tuesday directory has curlies.

PhilRunninger:
What version of Vim are you using? There have been several changes to the `glob()` and `globpath()` functions over the years (these functions are used to get a directories children), and I wonder if your particular version has not been taken into account. Please include the whole content of the `:version` command.

PhilRunninger:
I believe I've figured it where it's happening, but not necessarily why. Using this example,
```
$ mkdir /tmp/{foo}
$ touch /tmp/{foo}/.bar
```
Vim exhibits the following behavior
```vim
:echo globpath('/tmp/{foo}', '.*')
/tmp/{foo}/./
/tmp/{foo}/../
/tmp/{foo}/.bar
:echo globpath('/tmp/\{foo\}', '.*')
/tmp/{foo}/.
/tmp/{foo}/..
/tmp/{foo}/.bar
```
Notice the difference between `{foo}` and `\{foo\}`. I'm not sure what the special meaning of the curly braces is inside the `globpath()` function; I haven't been able to find anything in the help pages or on Google.

[...]

Issue #492, opened 2015-09-25
Not able to do anything with file named "[test].md" (m, o , enter do not work)


netei:
Whenever you name a file `[test].md`, it is not possible to do anything with it inside NerdTree (pressing m and any other keys has absolutely no effect)

This seems to come from the function UI._stripMarkup, specifically this line : https://github.com/scrooloose/nerdtree/blob/8c7534bc9e1d0c2f18bf94c1440c4ae6b102b5d6/lib/nerdtree/ui.vim#L359

```
function! s:UI._stripMarkup(line, removeLeadingSpaces)
    let line = a:line
    "remove the tree parts and the leading space
    let line = substitute (line, g:NERDTreeUI.MarkupReg(),"","")

    "strip off any read only flag
    let line = substitute (line, ' \[RO\]', "","")

    "strip off any bookmark flags
    let line = substitute (line, ' {[^}]*}', "","")

    "strip off any executable flags
    let line = substitute (line, '*\ze\($\| \)', "","")

    "strip off any generic flags
    let line = substitute (line, '\[[^]]*\]', "","")

    let wasdir = 0
    if line =~# '/$'
        let wasdir = 1
    endif
    let line = substitute (line,' -> .*',"","") " remove link to
    if wasdir ==# 1
        let line = substitute (line, '/\?$', '/', "")
    endif

    if a:removeLeadingSpaces
        let line = substitute (line, '^ *', '', '')
    endif

    return line
endfunction
```

I guess the changes that should be made to make this work would be huge (because the filename should not be extracted from the view layer, but from an other layer)

lifecrisis:
This issue is directly related to #680. It is a confirmed bug. I independently discovered the problem with `_stripMarkup()` previously, and this user has reached the same conclusion.

PhilRunninger:
@netei , would you be willing to test the pull request #868 I've put together for this issue? Details are in the PR. Thanks.

PhilRunninger:
Fixed in #868.

Issue #469, opened 2015-07-24
Сan't be used file if it name start with "space".


megosparta:
Can't open nerdtree menu ('m' key) under files with names started with whitespace.

PhilRunninger:
@megosparta , would you be willing to test the pull request #868 I've put together for this issue? Details are in the PR. Thanks.

PhilRunninger:
Fixed in #868.

PR #868, opened 2018-08-04
Support unusual characters in file and directory names

PhilRunninger:
Fixes: #680 #492 #469 #424 #391 #217 

This PR has the potential to fix a bunch of issues. The problem with these issues is related to parsing the file or directory name from the current line of text. In short, I fixed it by surrounding the filename with non-breaking spaces (ASCII code 160), and using that pair of delimiters to find the filename in the text when needed. I also added syntax highlighting to conceal this character, making it appear as if nothing has been added. The advantages to using this character are:

1. It is **very** unlikely to be used in a filename.
1. If `conceal` is not compiled into your Vim, the only side effect is that the tree and trailing flags are pushed over by one extra "space".

There may be cases where this doesn't work so well, though, and some improvements will need to be made to this PR. The one I can think of right away is editing in an environment that doesn't support Unicode characters, or whatever the non-breaking space is. `ssh` comes to mind as a possible trouble spot.

`conceal` was added in 7.3, but with the `if has("conceal")` in the syntax file, we should be ok. It will just have the added spacing indicated above.

I would like to enlist the authors of the related issues to test this fully. Areas I'd like them and others to try are:
- [ ] scenarios listed in the issues
- [ ] gvim
- [ ] console vim
- [ ] using the mouse to interact with NERDTree
- [ ] possible side effects caused to other plugins like
    - [ ] nerdtree-git-plugin
    - [ ] devicons
- [ ] ssh
- [ ] older versions of vim 7.0 - 7.2
- [ ] Windows
- [ ] Linux
- [ ] Mac

Of course, I don't expect everyone to test all these areas, but given enough people we can likely cover them all.

### DIFF
--- a/lib/nerdtree/path.vim
+++ b/lib/nerdtree/path.vim
@@ -42,7 +42,7 @@ function! s:Path.cacheDisplayString() abort
     let self.cachedDisplayString = g:NERDTreeNodeDelimiter . self.getLastPathComponent(1)
 
     if self.isExecutable
-        let self.cachedDisplayString .= (self.cachedDisplayString =~ g:NERDTreeNodeDelimiter.'$' ? '' : g:NERDTreeNodeDelimiter) . '*'
+        let self.cachedDisplayString = self.addDelimiter(self.cachedDisplayString) . '*'
     endif
 
     let self._bookmarkNames = []
@@ -52,15 +52,24 @@ function! s:Path.cacheDisplayString() abort
         endif
     endfor
     if !empty(self._bookmarkNames) && g:NERDTreeMarkBookmarks == 1
-        let self.cachedDisplayString .= (self.cachedDisplayString =~ g:NERDTreeNodeDelimiter.'$' ? '' : g:NERDTreeNodeDelimiter) . ' {' . join(self._bookmarkNames) . '}'
+        let self.cachedDisplayString = self.addDelimiter(self.cachedDisplayString) . ' {' . join(self._bookmarkNames) . '}'
     endif
 
     if self.isSymLink
-        let self.cachedDisplayString .= (self.cachedDisplayString =~ g:NERDTreeNodeDelimiter.'$' ? '' : g:NERDTreeNodeDelimiter) . ' -> ' . self.symLinkDest
+        let self.cachedDisplayString = self.addDelimiter(self.cachedDisplayString) . ' -> ' . self.symLinkDest
     endif
 
     if self.isReadOnly
-        let self.cachedDisplayString .= (self.cachedDisplayString =~ g:NERDTreeNodeDelimiter.'$' ? '' : g:NERDTreeNodeDelimiter) . ' ['.g:NERDTreeGlyphReadOnly.']'
+        let self.cachedDisplayString = self.addDelimiter(self.cachedDisplayString) . ' ['.g:NERDTreeGlyphReadOnly.']'
+    endif
+endfunction
+
+" FUNCTION: Path.addDelimiter() {{{1
+function! s:Path.addDelimiter(line)
+    if a:line =~# '\(.*' . g:NERDTreeNodeDelimiter . '\)\{2}'
+        return a:line
+    else
+        return a:line . g:NERDTreeNodeDelimiter
     endif
 endfunction
 
--- a/lib/nerdtree/path.vim
+++ b/lib/nerdtree/path.vim
@@ -42,7 +42,7 @@ function! s:Path.cacheDisplayString() abort
     let self.cachedDisplayString = g:NERDTreeNodeDelimiter . self.getLastPathComponent(1) . g:NERDTreeNodeDelimiter
 
     if self.isExecutable
-        let self.cachedDisplayString = self.cachedDisplayString . '*'
+        let self.cachedDisplayString .= '*'
     endif
 
     let self._bookmarkNames = []
--- a/lib/nerdtree/path.vim
+++ b/lib/nerdtree/path.vim
@@ -43,10 +43,10 @@ endfunction
 
 " FUNCTION: Path.cacheDisplayString() {{{1
 function! s:Path.cacheDisplayString() abort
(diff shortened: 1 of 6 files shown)


lifecrisis:
Wow, Phil!  You've been busy.  I'll make sure to take a look at this this week!

PhilRunninger:
@lifecrisis , did you get a chance to review this yet? I'm going to start tagging the issue authors for their help in reviewing this too.

jsdevel:
So far, doesn't seem to have fixed #555 

not-working

xcambar:
It fixes #680 🎉

jsdevel:
@PhilRunninger fixes #555 !

works

lifecrisis:
It seems that the functionality is all there!  Awesome!

I still want to take a closer look at the code itself, so I'll try to make it happen later this week.  This is a pretty *high-impact* change!

lifecrisis:
I'm getting hung up on something small here...

Go to the line of any NERDTree node.  Press `0` and then press `l` several times.  You'll see that the `l` motion will "catch" on the first word of the directory.  Any way to avoid this problem while still concealing the characters?  Nothing jumps out at me immediately.

lifecrisis:
I also noticed that, now, I can move the cursor beyond the end of the line.  I.e., I can move the cursor to the trailing nbsp character.

I think I have a solution to this...

Instead of relying on concealing the character, just show the first nbsp character and only use the second one if there are trailing flags (by having the process for appending the flags also add the nbsp first).

This would mean that a node would be delimited like `[nbsp][file name][nbsp/eol][flags]`.  This would get rid of the change to how the cursor moves around.

Also, just curious, how are we determining that nbsp is even available in the user's context?  What is the likelihood that this character will be available?  Is it almost certain?

PhilRunninger:
@lifecrisis , I tried implementing your solution, but I think I'll stick with what's committed already. On filenames and single directories, there are no issues, and I can parse them easily. Here are some examples: (A colon is used in place of the nbsp below.)
```
:.vim/
:nerdtree/: {nerdtree}
:path.vim
: space_in_first_position.txt
:vimrc: {vimrc}
:foobar: -> /tmp/foobar
```
However, when dealing with cascaded directories, we have this situation.
```
:lib/:nerdtree/
:foo/:bar/: {foobar}
```
There is no easy way to distinguish the intermediate `:` from a trailing one, and as a result, files in these directories cannot be opened.

I think the cursor's "catching" on the leading nbsp, and moving beyond the node where the trailing nbsp is, are minor issues, as horizontal movement isn't really needed (and probably uncommon) in NERDTree.

Whether nbsp is available in all fonts is a mystery. I have no way of knowing for sure, but I'd guess it fairly likely is. That being said, since it's being concealed anyway (concealing requires 7.3+), it shouldn't really matter how that character is rendered.

lifecrisis:
Ever thought about just associating each node with the line number it's written on though a Vim dictionary?  Then we could just render each node as we please.  The dictionary could be updated every time the NERDTree is drawn. Is this crazy?  I'd be worried that that process might become slow for large directories.

PhilRunninger:
I like that idea. It would take some time to get everything right, and performance could be a problem, but we won't know if we don't try. What do you suggest?
1. Adding line number to the node objects in `tree_file_node.vim` and `tree_dir_node.vim`, or
1. A new separate dictionary/array associating line number to filename, or
1. Something else

lifecrisis:
It's a pretty complicated idea... I'll have to think about it for a bit.

lifecrisis:
I thought about what you wrote above:

> There is no easy way to distinguish the intermediate : from a trailing one, and as a result, files
> in these directories cannot be opened.

I'm not sure why you need intermediate nbsp characters... the NERDTree directory separator is always the `/` character.  Furthermore, you always know that the trailing nbsp character is the trailing one because it's the last one.

I'm not really following the thinking here.  Sorry if I'm just confusing myself...

PhilRunninger:
Assuming we use the trailing nbsp only when adding the `*`, `{bookmarks}`, `-> symlink`, or `[RO]` tags,...

The nodes are separate objects in memory, so each dir node would have the leading delimiter.
```
:lib/
  :nerdtree/
```
but when they are combined for cascading, they look like this 
```
:lib/:nerdtree/
```
This is the behavior you noticed in this [comment](https://github.com/scrooloose/nerdtree/pull/868#issuecomment-415976611) above. (Remember, at that time it would have been `:lib/::nerdtree/:`.) In that case, it was treating intermediate delimiters as part of the directory name, until I made the fix in e1af7a314bb7877d11d758953320aa959004eb68.

Getting rid of the leading nbsp and everything before it is easy, but if we blindly get rid of the last nbsp and everything after it, we could end up losing the last folder of this cascaded directory node.

`:lib/:nerdtree/` erroneously becomes `lib/`, while 
`:lib/:nerdtree/: {bookmark}` correctly becomes `lib/nerdtree/`.

So it's a specific case of cascaded directories and no tags that causes the suggested fix to fail.



I just thought of a way to make this work. The code that combines the parent and child directory nodes can strip the leading nbsp off each child directory. I haven't looked at the code to see if it will work, but I think that's the way to go.

lifecrisis:
> I just thought of a way to make this work. The code that combines the parent and child directory nodes can strip the leading nbsp off each child directory. I haven't looked at the code to see if it will work, but I think that's the way to go.

That's exactly what I meant!  I feel like I'm being a bit difficult on this issue, but I'm adamant that this is something we need to get right.  It's such a fundamental change, and it would be very useful.

(Nice GIF by the way.)

Perhaps we could go with node ID numbers, associative arrays, etc., in the next version of the NERDTree and just use this for now (you're only affecting a few dozen lines of code here).

PhilRunninger:
> I feel like I'm being a bit difficult on this issue, but ...

No worries. I'm glad to have you reviewing this. It results in a much better solution that way. New commits (final ones, I hope) coming soon.

PR #878, opened 2018-08-22
NERDTreeCWD: reset CWD if changed by NERDTreeFocus

### LINKED ISSUE #855: NERDTreeCWD and autochdir do not cooperate
Not sure if this is a bug or not, `NERDTreeCWD` appears to change directory to the `:pwd` of the `NERDTree`'s buffer and not to the current buffer's when using `autochdir`.

### Environment

* OS: macOS High Sierra
* Editor: MacVim
* Version: 8.1
* NERDTree version: d6032c8

### Process

1. `mvim ~/`
2. `:set autochdir`
3. `:cd ~/test/`
4. `:NERDTreeCWD`

### Current Result

NERDTree's root is the `~/`.

### Expected Result

NERDTree's root should be the `~/test/`.

It works when changing the second step (2.) to `:set noautochdir`.

PhilRunninger:
Fixes #855. 

When the user has `'autochdir'` turned on, opening a new NERDTree will
cause the current working directory to change. To prevent this
happening, remember the CWD and reset it if NERDTreeFocus caused it to
change.

### DIFF
--- a/plugin/NERD_tree.vim
+++ b/plugin/NERD_tree.vim
@@ -202,7 +202,11 @@ function! NERDTreeFocus()
 endfunction
 
 function! NERDTreeCWD()
+    let l:cwd = getcwd()
     call NERDTreeFocus()
+    if l:cwd != getcwd()
+        exec 'cd '.l:cwd
+    endif
     call nerdtree#ui_glue#chRootCwd()
 endfunction
 
--- a/plugin/NERD_tree.vim
+++ b/plugin/NERD_tree.vim
@@ -203,6 +203,11 @@ endfunction
 
 function! NERDTreeCWD()
 
+    if empty(getcwd())
+        call nerdtree#echoWarning('current directory does not exist')
+        return
+    endif
+
     try
         let l:cwdPath = g:NERDTreePath.New(getcwd())
     catch /^NERDTree.InvalidArgumentsError/
--- a/plugin/NERD_tree.vim
+++ b/plugin/NERD_tree.vim
@@ -202,12 +202,15 @@ function! NERDTreeFocus()
 endfunction
 
 function! NERDTreeCWD()
-    let l:cwd = getcwd()
+    let l:cwdPath = g:NERDTreePath.New(getcwd())
     call NERDTreeFocus()
-    if l:cwd != getcwd()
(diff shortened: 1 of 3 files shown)


plugin/NERD_tree.vim
```diff
@@ -202,7 +202,11 @@ function! NERDTreeFocus()
 endfunction
 
 function! NERDTreeCWD()
+    let l:cwd = getcwd()
     call NERDTreeFocus()
+    if l:cwd != getcwd()
+        exec 'cd '.l:cwd
```
lifecrisis: You should wrap `l:cwd` with `fnameescape()` here.

lifecrisis:
If you read the documentation at `:help :NERDTreeCWD`, it leads one to believe that our function should be more like this:

```viml
function! NERDTreeCWD()
    let l:cwdPath = g:NERDTreePath.New(getcwd())
    call NERDTreeFocus()

    if b:NERDTree.root.path.equals(l:cwdPath)
        return
    endif

    let l:newRoot = g:NERDTreeFileNode.New(l:cwdPath, b:NERDTree)
    call b:NERDTree.changeRoot(l:newRoot)
endfunction
```

This matches the documentation perfectly, reproduced below:

```
:NERDTreeCWD                                                  :NERDTreeCWD
    Change tree root to current directory. If no NERD tree exists for this
    tab, a new tree will be opened.
```

I assert that changing the working directory in any way is not in the scope of this command!

Also, the documentation may be improved a tad...

```
:NERDTreeCWD                                                  :NERDTreeCWD
    Change the NERDTree root to the current working directory.  If no
    NERDTree exists for this tab, a new one is opened.
```

lifecrisis:
I thought it more last night...  I think my above comment and suggested function definition is right on the money.  Even in the documentation for `'autochdir'`, we see:

```
Note: When this option is on some plugins may not work.
```

I just don't think the setting is working the way the user things it should.

lifecrisis:
Hey @PhilRunninger.  Updates look great!

Just added a couple of commits on top of yours.  If you're okay with everything, then so am I.  Let me know if this is good or if you think of anything else.

PhilRunninger:
Looks good. I hadn't considered the `CD` mapping. It's good to have them use the same code.

PR #886, opened 2018-09-13
Fix the <C-J>/<C-K> motions, which currently fail with cascades

lifecrisis:


lifecrisis:
**NOTE:**  The last two commits I added fix the `J` and `K` default mappings, which suffer from a similar issue.  This is not to big of a problem, but it's quite annoying when you do run into it!

lifecrisis:
My favorite type of PR is one in which I can fix an bug *and* remove more code than I add.  I think this is a good one.

PR #887, opened 2018-09-15
Use a better arg for FINDSTR when using the m,l command in Windows.

### LINKED ISSUE #883: folder listing (menu m-l) not working for my locale
### Problem
Hitting m-l on a folder in NERDTree gives me 'shell command failed'. I tracked down the error to
FINDSTR "^[012][0-9]/[0-3][0-9]/[12][0-9][0-9][0-9]"
in file fs_menu.vim, function NERDTreeListNodeWin32(). The system date format for my Windows installation is set to yyyy-MM-dd, whereas the regular expression searches for MM-dd-yyyy.

### Environment
Vim 8.1.1 with latest NERDTree, Windows 10 set to English language

PhilRunninger:
Fixes #883.

The only reason for the `FINDSTR` is to remove the extra seven lines of boilerplate that the `DIR` command thinks you need to know. Instead of "grepping" for the date (in its many different formats), we can more easily look for the filename itself.

### DIFF
--- a/lib/nerdtree/ui.vim
+++ b/lib/nerdtree/ui.vim
@@ -194,51 +194,52 @@ function! s:UI.getPath(ln)
     return toReturn
 endfunction
 
-" FUNCTION: s:UI.getLineNum(file_node){{{1
-" returns the line number this node is rendered on, or -1 if it isnt rendered
-function! s:UI.getLineNum(file_node)
-    " if the node is the root then return the root line no.
-    if a:file_node.isRoot()
+" FUNCTION: s:UI.getLineNum(node) {{{1
+" Return the line number where the given node is rendered.  Return -1 if the
+" given node is not visible.
+function! s:UI.getLineNum(node)
+
+    if a:node.isRoot()
         return self.getRootLineNum()
     endif
 
-    let totalLines = line("$")
+    let l:pathComponents = [substitute(self.nerdtree.root.path.str({'format': 'UI'}), '/\s*$', '', '')]
+    let l:currentPathComponent = 1
 
-    " the path components we have matched so far
-    let pathcomponents = [substitute(self.nerdtree.root.path.str({'format': 'UI'}), '/ *$', '', '')]
-    " the index of the component we are searching for
-    let curPathComponent = 1
+    let l:fullPath = a:node.path.str({'format': 'UI'})
 
-    let fullpath = a:file_node.path.str({'format': 'UI'})
+    for l:lineNumber in range(self.getRootLineNum() + 1, line('$'))
+        let l:currentLine = getline(l:lineNumber)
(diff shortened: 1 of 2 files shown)


PhilRunninger:
I forgot to check directories. To be honest, what shows in `master` isn't all that informative either. It'll likely require an `if directory ... else ... endif` construct, and more than a simple `DIR` command. Let me play with it a bit more.

PhilRunninger:
Taking what we get from `DIR /q macros`:
```
 Volume in drive C has no label.
 Volume Serial Number is 3C81-BABD

 Directory of C:\Program Files (x86)\Vim\vim81\macros

05/26/2018  04:25 PM    <DIR>          BUILTIN\Administrators .
05/26/2018  04:25 PM    <DIR>          BUILTIN\Administrators ..
03/28/2016  02:56 PM               124 BUILTIN\Administrators editexisting.vim
05/26/2018  04:25 PM    <DIR>          BUILTIN\Administrators hanoi
03/28/2016  02:44 PM               109 BUILTIN\Administrators justify.vim
05/18/2012  07:38 AM               288 BUILTIN\Administrators less.bat
05/18/2012  07:37 AM               588 BUILTIN\Administrators less.sh
03/31/2017  02:47 PM             5,567 BUILTIN\Administrators less.vim
05/26/2018  04:25 PM    <DIR>          BUILTIN\Administrators life
02/09/2017  03:16 PM               124 BUILTIN\Administrators matchit.vim
05/26/2018  04:25 PM    <DIR>          BUILTIN\Administrators maze
03/28/2016  04:03 PM             1,120 BUILTIN\Administrators README.txt
03/28/2016  03:16 PM               115 BUILTIN\Administrators shellmenu.vim
03/28/2016  03:07 PM               114 BUILTIN\Administrators swapmous.vim
05/26/2018  04:25 PM    <DIR>          BUILTIN\Administrators urm
               9 File(s)          8,149 bytes
               6 Dir(s)  21,076,803,584 bytes free
```
what do you think of this transformation?
```
05/26/2018  04:25 PM  <DIR>  BUILTIN\Administrators  macros  [9 File(s) 8,149 bytes,  4 Dir(s)]
```

lifecrisis:
That looks fine to me.  Can we be sure that the output of this command will remain stable?  I'm not so familiar with Windows utilities.

PhilRunninger:
@lifecrisis, I decided to go with the built-in Vim functions, `getfsize`, `getftime`, and so on, to find the same information. The only thing I don't like is the size of 0 for directories, but that's a Windows thing. It's similar to what we currently see in the `master` branch where size is left blank.

Unfortunately, there is no Vim function equivalent to the `DIR` command's `/Q` option, which displays the owner of the file.

PR #892, opened 2018-10-08
Fix typo in help file

lvoisin:
Remove duplicated word.


PhilRunninger:
Good catch! Thanks for the contribution.

PR #898, opened 2018-10-17
Force sort to recalculate the cached sortKey.

### LINKED ISSUE #880: Sorting is not refreshed after the directory is renamed
### Environment

* Operating System:  Windows10
* Vim version `:version`:  Vim 8.1
* NERDTree version `git rev-parse --short HEAD`:  808f5b2
* NERDTree settings applied in your vimrc, if any: 
    ```vim
    let g:NERDTreeNaturalSort = 1
    ```

### Process

1. Just mkdir like follow:
default
2. Rename dir `01` to `03`, you will get:
default

### Expected Result

Sorting need be refreshed after the directory is renamed.

PhilRunninger:
Fixes #880.

The problem in issue #880 was caused by the sort using the old sortKey.
For example, given nodes A, B, and C, if B were renamed to D, the sort
was still using B as its sortKey, thus not moving it.

It's a bit of a hack, but if we set g:NERDTreeOldSortOrder to an empty
list, the cached sortKey will be recalculated. I did the same thing for
both the Copy and Add functions as well.

### DIFF
--- a/nerdtree_plugin/fs_menu.vim
+++ b/nerdtree_plugin/fs_menu.vim
@@ -128,6 +128,8 @@ function! NERDTreeAddNode()
         let parentNode = b:NERDTree.root.findNode(newPath.getParent())
 
         let newTreeNode = g:NERDTreeFileNode.New(newPath, b:NERDTree)
+        " Emptying g:NERDTreeOldSortOrder forces the sort to
+        " recalculate the cached sortKey so nodes sort correctly.
         let g:NERDTreeOldSortOrder = []
         if empty(parentNode)
             call b:NERDTree.root.refresh()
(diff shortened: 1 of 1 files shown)


PhilRunninger:
The most logical place to DRY this code up is to put `let g:NERDTreeOldSortOrder = []` in the `b:NERDTree.root.refresh()` function before sorting, but we would run into the same performance issue you noticed in #856. Question is: How often will refresh be called in the real world?

lifecrisis:
> Question is: How often will refresh be called in the real world?

I wonder this myself.  I use it all the time... but it's mainly when I'm testing NERDTree code!

In light of your observation, I say merge it.  The `fs_menu.vim` code needs some refactoring later, anyway.  A comment may help clarify things, but I'm otherwise okay with everything you've got here.

Issue #90, opened 2011-09-14
sort by modification date


XedMada:
would be very handy for some uses. e.g. notational velocity replacement.

tebeka:
+1

xinxingzhao:
+1

hiberabyss:
+1

AndrewRadev:
I created a fork of the plugin that enables sorting by access time: https://github.com/AndrewRadev/nerdtree/tree/sort-by-atime. The instructions on how to enable it and toggle it on or off, are at the top of the README.

I'm not making a PR, because it's a bit hacky. If Scrooloose wanted to implement the feature, I imagine he'd make a setting for a sorting callback or something, but I just toggle it with a global variable. It's also slow for large directories, since every single file needs to get `stat`-ed to get the access time, which means an extra shell-out. But, depending on your use case, you might find it useful.

tadassce:
@AndrewRadev it's exactly what I needed, thanks!!!

rizjoj:
+1

xbot:
+1

seanw2020:
Here's my workaround for NERDTree's lack of "timestamp sort":

`:Ex /some/path/` (opens netrw in vertical split. Alternatively, `:Se` for horizontal split)
`s` (to sort by time)
`r` (reverse order -- look for `"Sorted by time reversed"`)
`Enter` (open your file)
`:NerdTreeFind` (navigates NERDTree to the currently open file)

PhilRunninger:
Implemented by #901

leonschreuder:
This implementation is great, but not very convenient. I've written this workaround:
```viml
command! NERDTreeSortTimeToggle call NERDTreeSortTimeToggle()

fu! NERDTreeSortTimeToggle()
  if !exists('g:NERDTreePrevSortOrder')
    let g:NERDTreePrevSortOrder = copy(g:NERDTreeSortOrder)
    let g:NERDTreeSortOrder += ['[[-timestamp]]']
    call nerdtree#ui_glue#invokeKeyMap("R")
  else
    let g:NERDTreeSortOrder = g:NERDTreePrevSortOrder
    unlet g:NERDTreePrevSortOrder
    call nerdtree#ui_glue#invokeKeyMap("R")
  endif
endfu
```
but would prefer something built-in.

PR #902, opened 2018-10-22
Sort by modification date

PhilRunninger:
Implements #90.

See commit messages for details.


PhilRunninger:
Closing PR. It was a duplicate that got made when github.com was struggling to become functional again after an outage. At that time #901 was returning a 404 error, and then this one was.

PR #903, opened 2018-10-24
Fix vsplit to not open empty buffer

### LINKED ISSUE #616: Nerdtree Split opens empty Buffers
Hi,

I'm not exactly sure when this happens but sometimes when I open new Files via [s] in a split view Nerdtree seems to create an empty buffer additionally - or is this Vim? 

Basically I have Nerdtree (toggeled) and one split with one buffer. After choosing a second file and pressing [s] another split happens with the new file inside but a new empty buffer appears in the :buffers list. I recognized this when closing buffers.

It's somehow not happening if I split open buffers via Ctrlp.

Any ideas?

AwkwardKore:
Fixes #616.

Steps to reproduce bug:
1. Open a file from nerdtree
2. Delete buffer with :bd
3. Reopen same file from nerdtree with vsplit (s).

This reopens the buffer that didn't unload from memory with :bd and creates a new empty buffer with [No name].

### DIFF
--- a/lib/nerdtree/opener.vim
+++ b/lib/nerdtree/opener.vim
@@ -220,7 +220,7 @@ function! s:Opener._newVSplit()
     endif
 
     call nerdtree#exec('wincmd p')
-    vnew
+    vsplit
 
     let l:currentWindowNumber = winnr()
 


PhilRunninger:
Thank you for your contribution; however, I cannot accept this change. I don't see any difference in the way NERDTree behaves between the current master and your branch. I also don't believe that your symptoms describe a problem that needs fixing.

Vim always needs a buffer to show. When you delete the only buffer using the `:bd` command, Vim needs to create a new buffer to accept new input. If you were then to open an existing file from NERDTree, using the `o` command, Vim replaces that `[No Name]` buffer if it's still unchanged. On the other hand, if you use NERDTree's `s` command to split the window, the `[No Name]` buffer must remain in case the user decides to put text in that window. 

I believe Vim and NERDTree are acting as expected.

Feel free to defend your proposed changes by opening a new issue with a detailed description of the problem; include your NERDTree settings in your vimrc and the exact steps needed to repeat the problem.

And feel free to contribute to PRs and issue in the future.

PR #923, opened 2018-11-26
Allow modification of menu keybindings

Leandros:
My keyboard layout has different keys on 'j' & 'k'. This allows changing the defaults.

### DIFF
--- a/lib/nerdtree/menu_controller.vim
+++ b/lib/nerdtree/menu_controller.vim
@@ -49,7 +49,7 @@ endfunction
 
 "FUNCTION: MenuController._echoPrompt() {{{1
 function! s:MenuController._echoPrompt()
-    echo "NERDTree Menu. Use j/k/enter and the shortcuts indicated"
+    echo "NERDTree Menu. Use " . g:NERDTreeMenuDown . "/" . g:NERDTreeMenuUp . "/enter and the shortcuts indicated"
     echo "=========================================================="
 
     for i in range(0, len(self.menuItems)-1)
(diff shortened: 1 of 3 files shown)


lib/nerdtree/menu_controller.vim
```diff
@@ -71,9 +71,9 @@ endfunction
 "change the selection (if appropriate) and return 1 if the user has made
 "their choice, 0 otherwise
 function! s:MenuController._handleKeypress(key)
-    if a:key == 'j'
+    if a:key == 'j' || a:key == g:NERDTreeMenuDown
```
PhilRunninger: Change this to
```vim
if a:key == g:NERDTreeMenuDown
```
and initialize the variable in this section: https://github.com/scrooloose/nerdtree/blob/master/plugin/NERD_tree.vim#L116-L150

lib/nerdtree/menu_controller.vim
```diff
@@ -71,9 +71,9 @@ endfunction
...
         call self._cursorDown()
-    elseif a:key == 'k'
+    elseif a:key == 'k' || a:key == g:NERDTreeMenuUp
```
PhilRunninger: Same change here.

Leandros:
Reasonable. Will add the changes later today.

Leandros:
Incorporated your suggestions. Although, I didn't update the `j/k` text, since by default you'd still use `j` and `k` for navigation. The variables are documented in NERDTree.txt.

PhilRunninger:
That text needs to change. The default is `j/k`, but if you override them, that needs to be reflected in the text.
```vim
    echo "NERDTree Menu. Use " . g:NERDTreeMenuDown . "/" . g:NERDTreeMenuUp . "/enter and the shortcuts indicated"
```

Leandros:
Ahh, I didn't think of that. Makes sense.

Leandros:
Fixed.

PhilRunninger:
Excellent! Thanks for the contribution, @Leandros.

Leandros:
Thank you for NERDTree!

Issue #895, opened 2018-10-11
Folders and files improperly rendered


PhilRunninger:
### Environment

* Operating System: **Windows 10**
* Vim version `:version`: **8.1**
* NERDTree version `git rev-parse --short HEAD`: **Pull Request #890**
* NERDTree settings applied in your vimrc, if any:
    ```vim
    ```

### Process

Create a new drive letter if needed, `G:` in this case.
```
C:\> subst G: C:\users\Name\Documents
```
Then in vim, as described in #860,
1. `:set shellslash`
1. `:cd g:/`
1. `:NERDTree`

### Current Result

With the changes in PR #890 and the steps outlined above, NERDTree doesn't render the files and folders correctly, nor does it open them. See the screenshots in #860.

### Expected Result

Files and folders opening and closing like normal.

PhilRunninger:
Fixed by #935.

PR #890, opened 2018-09-22
Fix process for creating path for a drive letter on Windows

lifecrisis:
See issue #860.

Issue #860, opened 2018-07-16
`shellslash` option interference nerdtree


zgpio:
### Environment

* Operating System: win10
* Vim version `:version`:  8.1
* NERDTree version `git rev-parse --short HEAD`: latest 
* NERDTree settings applied in your vimrc, if any:
    ```vim
    " no need
    ```

### Process

1.  copy `nerdtree` repo to `...\vimfiles\pack\github\start`
2. start vim, run `:set shellslash` and `:cd d:/`
3. run `:NERDTree`, you will get message
```
NERDTree: No bookmark or dirctory found for:
```

lifecrisis:
Confirmed this bug.  It will take some fixing and a little clean-up to get this working properly.  Thanks.

zgpio:
Thanks a lot.

zgpio:
And `copy the current node` can not work.

zgpio:
Yes, the original problem is fixed, but the `shellslash` option still interference nerdtree.

zgpio:
with the patch:
start vim, run `:set shellslash` and `:cd d:/`, and then expand a directory, you will find like that:
![image](https://user-images.githubusercontent.com/19503791/45923649-54f3fb80-bf1e-11e8-9361-3f4643c25bf8.png)

lifecrisis:
What kind of files are in that directory?  Are they all links?  Please provide more information on how to make this happen in my Vim instance.

zgpio:
Sorry for that.
with the patch 860:
start vim, run `:set shellslash` and `:cd d:/`, and then expand a directory, you will find like that(fig1):
![image](https://user-images.githubusercontent.com/19503791/45934654-b2964f80-bfd3-11e8-9b4a-fbb46a9a0fdd.png)
but it should be fig2:
![image](https://user-images.githubusercontent.com/19503791/45934670-0e60d880-bfd4-11e8-8653-d36954039296.png)

If set noshellslash, vim will block when I expand a directory, so that I waited a long time to list the directories(fig2).

PhilRunninger:
@zgpio , PR #935 has just been merged into master. Please give it a try and close this issue it it's been resolved.

zgpio:
Sorry, replying so late.
Now it seems that nerdtree is working correctly.

PR #935, opened 2019-01-18
fixed root path error for windows

zcodes:
there are two problems when change cwd to windows disk root like 'd:\\'

1. if set 'noshellslash', nerdtree can open, but will crash when open sub directory of 'd:\\', such as 'd:\\docs'
2. if set 'shellslash', nerdtree can be opened, and error message will echo `NERDTree: No bookmark or dirctory found for:`

the core point of these problems is that the function `getcwd()` result diffrent style for 'd:\\' and 'd:\\docs',
if set 'noshellslash', getcwd() return 'd:\\' with trailing slash and 'd:\\docs' without trailing slash.
if set 'shellslash', getcwd() return 'd:/' and 'd:/docs', the function s:Path.AbsolutePathFor() does not handle
these differences properly.
 
this pr maybe related to #860 #890 #895

### DIFF
--- a/lib/nerdtree/path.vim
+++ b/lib/nerdtree/path.vim
@@ -15,7 +15,7 @@ function! s:Path.AbsolutePathFor(pathStr)
     let l:prependWorkingDir = 0
 
     if nerdtree#runningWindows()
-        let l:prependWorkingDir = a:pathStr !~# '^.:\(\\\|\/\)' && a:pathStr !~# '^\(\\\\\|\/\/\)'
+        let l:prependWorkingDir = a:pathStr !~# '^.:\(\\\|\/\)\?' && a:pathStr !~# '^\(\\\\\|\/\/\)'
     else
         let l:prependWorkingDir = a:pathStr !~# '^/'
     endif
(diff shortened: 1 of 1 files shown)


PhilRunninger:
@zcodes, Thanks for the contribution. I have no way to verify the proposed changes for about a week. In the meantime, I have some comments on a couple lines of code.

PhilRunninger:
Dang! FastHub (my phone app for GitHub) won't let me make comments on specific lines, so...  Will line 28 work if you use `s:Path.Slash()` and eliminate the OR condition? If so, I think you don't even need the `runningWindows()` check anymore.

PhilRunninger:
I believe the change to line 18 is not needed. That line is checking for the presence of one of the following: `C:\`, `C:/`, `\\`, or `//` at the **beginning** of the line.

zcodes:
for the line 28, I simplify the code.

for line 18, the change is needed, some condition, the args (the path string) passed to the function is just 'd:', not 'd:\\' or 'd:/', the result string will be 'd:d:/', that's why #860

PR #938, opened 2019-01-21
Add a NERDTreeMinimalMenu feature

tuzz:
Hello, thanks for maintaining a plugin I use all the time. ❤️ 

One thing I've wanted for a while is a more compact menu that reduces the amount of scrolling in Vim. I think a large portion of people who use NERDTree are already familiar with each of the menu items so this might be a nice addition to reduce noise.

Therefore, I've added a new configuration setting `NERDTreeMinimalMenu` for this. When this is set, a single-line menu is presented and the follow-up text for each action is presented as a single line, too.

Thanks

menu

### DIFF
--- a/doc/NERDTree.txt
+++ b/doc/NERDTree.txt
@@ -1126,15 +1126,13 @@ Default: 0
 
 This setting makes NERD tree use a smaller, more compact menu for adding,
 copying, deleting nodes. This menu fits on a single line so Vim doesn't need to
-scroll down to present it. Only the shortcut keys are displayed so this setting
-is recommended for users already familiar with the menu items.
+scroll down to present it. This setting is recommended for users already
+familiar with the menu items. It will look similar to this:
 
-It will look similar to this:
+  Menu: [ (a)dd ,m,d,r,o,q,c,l] (Use j/k/enter or shortcut):
 
-  Which action? (a*/m/d/r/o/q/c/l):
-
-An action can be selected with its shortcut key or by moving the position of the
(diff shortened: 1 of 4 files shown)


PhilRunninger:
I like this idea. I'm almost tempted to suggest this be the only way to show the menu.

I just merged a PR that caused a conflict to appear in yours. It won't take too much work to resolve it. While you're at it, what do you think of this suggestion? Instead of, or in addition to, an asterisk to mark the current selection, show its full text, something like one of these:
```
Menu: [(a)dd node /m/d/r/o/q/c/l] (Use j/k/enter or shortcut):
```
```
Menu: [a, (m)ove node ,d,r,o,q,c,l] (Use j/k/enter or shortcut):
```
```
Menu: [a|m| (d)elete node |r|o|q|c|l] (Use j/k/enter or shortcut):
```
I think it's important to make sure the `j/k/enter` option for navigating the menu is visible for those unfamiliar with it. Believe it or not, I didn't notice it until recently myself; I always just used the shortcut.

tuzz:
Hi Phil, thanks for the feedback. I've rebased and updated the menu in [this](https://github.com/tuzz/nerdtree/commit/3d8eaa06b8e929a997f9b34a78afbdf45bf78939) commit.

I played with it a bit and I think this might be the cleanest way to present it. See the commit message for my rationale. What do you think?

minimal-menu

Thanks

PhilRunninger:
Looks good, @tuzz. Thanks for the addition.

Issue #947, opened 2019-02-14
Remapping alt-\ to open the file in vertical split throws error


rbhanot4739:
### Environment

* Operating System: Ubuntu 16.04.5 LTS
* Vim version `:NVIM v0.3.2-208-g2841e89`:  
* NERDTree version `5.0.0`: 
* NERDTree settings applied in your vimrc, if any:
    ```vim
    ```

### Process

1. Define these custom key bindings 
```
let g:NERDTreeMapOpenVSplit = '<a-\>'
```

### Current Result

So inside the NERDTree window when I press `a-\` I get the following error.
```
E114: Missing quote: "\")                                                                                                                                        
E116: Invalid arguments for function nerdtree#ui_glue#invokeKeyMap 
```
### Expected Result

I have mapped same key-stroke in normal for creating vsplits and it works fine. So what am i missing here ?

PhilRunninger:
Fixed in #948.

rbhanot4739:
Thank you very much, works like charm !!

PR #948, opened 2019-02-14
Escape a backslash so it can be used in a key mapping.

### LINKED ISSUE #735: Bindings with backslash causes error
Running vim 8.0.851 on linux, newest nerdtree.
I'm trying to make '\' open the selected file in a new vertical split.
I got it working just fine with Backspace, but can't seem to get '\' working.

It actually registers the bind but merely crashes. The backslash seems to be perhaps escaping a quotation mark?

My bind in my .vimrc looks like this.
```
let NERDTreeMapOpenSplit='<BS>'
let NERDTreeMapOpenVSplit='\'
```

When '\' is pressed on on a file in nerdtree the following error occurs:

```
E114: Missing quote: "\")
E116: Invalid arguments for function nerdtree#ui_glue#invokeKeyMap
```

Is there an alternative way for me to make this bind, or is it not possible with the current interpretation?

PhilRunninger:
Fixes #735, #947.

### DIFF
--- a/lib/nerdtree/key_map.vim
+++ b/lib/nerdtree/key_map.vim
@@ -51,6 +51,7 @@ function! s:KeyMap.bind()
     else
         let keymapInvokeString = self.key
     endif
+    let keymapInvokeString = escape(keymapInvokeString, '\')
 
     let premap = self.key == "<LeftRelease>" ? " <LeftRelease>" : " "
 


und3rdg:
@PhilRunninger thank you for this fix, it was make my day ;-]

Issue #968, opened 2019-03-26
E704 when opening nerdtree or new directory


Coleej:
With the newest commit, when I toggle Nerdtree or navigate to a new directory, Vim throws this error:

E704: Funcref variable name must start with a capital: l:numChildrenCached

The directory is not populated with any other nodes. I reverted back to commit: '288669db1fa888947deaf715d7469233f476909f' and the problem disappeared. 

### Environment

* Operating System: Linux, both ubuntu 18.04 (4.18 kernel) and Arch (5.0.4 kernel)

* Vim version `:version`:
```VIM - Vi IMproved 8.1 (2018 May 18, compiled Nov 26 2018 20:52:35)
Included patches: 1-547
Compiled by cjohnson@mobile
Huge version with GTK2-GNOME GUI.  Features included (+) or not (-):
+acl               +clientserver      +diff              +folding           +libcall           -mouse_gpm         +packages          -ruby              +termguicolors     +visual            +xsmp_interact
+arabic            +clipboard         +digraphs          -footer            +linebreak         -mouse_jsbterm     +path_extra        +scrollbind        +terminal          +visualextra       +xterm_clipboard
+autocmd           +cmdline_compl     +dnd               +fork()            +lispindent        +mouse_netterm     -perl              +signs             +terminfo          +viminfo           -xterm_save
+autochdir         +cmdline_hist      -ebcdic            +gettext           +listcmds          +mouse_sgr         +persistent_undo   +smartindent       +termresponse      +vreplace
+autoservername    +cmdline_info      +emacs_tags        -hangul_input      +localmap          -mouse_sysmouse    +postscript        +startuptime       +textobjects       +wildignore
+balloon_eval      +comments          +eval              +iconv             -lua               +mouse_urxvt       +printer           +statusline        +timers            +wildmenu
+balloon_eval_term +conceal           +ex_extra          +insert_expand     +menu              +mouse_xterm       +profile           -sun_workshop      +title             +windows
+browse            +cryptv            +extra_search      +job               +mksession         +multi_byte        -python            +syntax            +toolbar           +writebackup
++builtin_terms    +cscope            +farsi             +jumplist          +modify_fname      +multi_lang        +python3/dyn       +tag_binary        +user_commands     +X11
+byte_offset       +cursorbind        +file_in_path      +keymap            +mouse             -mzscheme          +quickfix          +tag_old_static    +vartabs           -xfontset
[...]

PR #969, opened 2019-03-26
fix _initChildren function value set to numChildrenCached error

terryding77:
#968


PhilRunninger:
@terryding77 , thanks for catching this.

PR #993, opened 2019-05-09
Fix minimum vim compare

jxdabc:
Actually, v:version of vim 7.3 is 703, NOT 730. So,
'if v:version < 730' should be 'if v:version < 703'.

Signed-off-by: Trol <jiaoxiaodong@xiaomi.com>


PhilRunninger:
@jxdabc Thanks for catching this. I already merged another PR with the same fix.

PR #994, opened 2019-05-09
Change version check to look for 703 not 730

vhalis:
My Vim reports as 704 through `:echo v:version` when it is version 7.4, so this number should be 703 not 730.

I don't have an older version of vim to verify this behaviour is consistent across versions, though


PhilRunninger:
Thanks @vhalis . Good catch!

PR #1014, opened 2019-07-03
Update ui_glue.vim

lkebin:
Fix `<CR>` key map on the bookmark


PhilRunninger:
Guess who forgot to test his code change. Oops! Good catch! Can you add the following line to CHANGELOG.md below `#### 5.2...`, so it's included in the PR?
```
- ***.3**: Fix <CR> key map on the bookmark (lkebin) #1014
```
Thanks

PR #1026, opened 2019-08-08
Suppress events for intermediate window/tab/buffer changes

### LINKED ISSUE #588: BufLeave and BufEnter event not getting triggered when re-entering the same buffer from NERDTree
As described  [here](https://github.com/kshenoy/vim-signature/issues/3#issuecomment-222421977), if I open NERDTree and go back to the same file by pressing `<CR>` in NERDTree, it doesn't trigger the `BufLeave` and `BufEnter` autocmds. 

Is this intentional?

### LINKED ISSUE #684: NERDTree: Refreshing the root node. This could take a while...

PhilRunninger:
### Description of Changes
Closes #588   

Got rid of the line that always ignores the `BufEnter`, `BufLeave`, and `VimEnter` events. So now, no events are ignored, and this allows `BufEnter` to be fired when opening an already open file, fixing #588.

But not so fast.....

Closes #684 

The change above causes events to be fired way more often than necessary because of NERDTree's need to switch windows at various times. To solve this, a new parameter was added to `nerdtree#exec()` that can cause **all** events to be ignored. This is set to 1 for all the intermediate window switches, and then to 0 for the final window switch to the selected file. This change, which comprises many lines of code throughout NERDTree, fixes the ping-ponging that was mentioned in #684 but has, until now, been left unaddressed.

---
### New Version Info

- [x] Derive a new version number. Increment the:
    - [ ] `MAJOR` version when you make incompatible API changes
    - [ ] `MINOR` version when you add functionality in a backwards-compatible manner
    - [x] `PATCH` version when you make backwards-compatible bug fixes
- [x] Update [CHANGELOG.md](https://github.com/scrooloose/nerdtree/blob/master/CHANGELOG.md), following this format/example:
    ```
    #### MAJOR.MINOR...
    - **.PATCH**: PR Title (Author) [#PR Number](link to PR)

    #### 5.1...
    - **.1**: Update Changelog and create PR Template (PhilRunninger) [#1007](https://github.com/scrooloose/nerdtree/pull/1007)
    - **.0**: Too many changes for one patch...
    ```

### DIFF
--- a/lib/nerdtree/nerdtree.vim
+++ b/lib/nerdtree/nerdtree.vim
@@ -44,14 +44,14 @@ function! s:NERDTree.Close()
         let l:useWinId = exists('*win_getid') && exists('*win_gotoid')
 
         if winnr() == s:NERDTree.GetWinNum()
-            call nerdtree#exec("wincmd p")
+            call nerdtree#exec("wincmd p", 1)
             let l:activeBufOrWin = l:useWinId ? win_getid() : bufnr("")
-            call nerdtree#exec("wincmd p")
+            call nerdtree#exec("wincmd p", 1)
         else
             let l:activeBufOrWin = l:useWinId ? win_getid() : bufnr("")
         endif
 
-        call nerdtree#exec(s:NERDTree.GetWinNum() . " wincmd w")
+        call nerdtree#exec(s:NERDTree.GetWinNum() . " wincmd w", 1)
         close
         if l:useWinId
             call nerdtree#exec("call win_gotoid(" . l:activeBufOrWin . ")")
@@ -98,7 +98,7 @@ endfunction
 "Places the cursor in the nerd tree window
 function! s:NERDTree.CursorToTreeWin()
     call g:NERDTree.MustBeOpen()
-    call nerdtree#exec(g:NERDTree.GetWinNum() . "wincmd w")
+    call nerdtree#exec(g:NERDTree.GetWinNum() . "wincmd w", 1)
 endfunction
 
 " Function: s:NERDTree.ExistsForBuffer()   {{{1
--- a/lib/nerdtree/nerdtree.vim
+++ b/lib/nerdtree/nerdtree.vim
@@ -52,7 +52,7 @@ function! s:NERDTree.Close()
         endif
 
         call nerdtree#exec(s:NERDTree.GetWinNum() . " wincmd w", 1)
-        close
+        call nerdtree#exec("close", 1)
         if l:useWinId
             call nerdtree#exec("call win_gotoid(" . l:activeBufOrWin . ")",0)
         else
--- a/lib/nerdtree/nerdtree.vim
+++ b/lib/nerdtree/nerdtree.vim
@@ -54,9 +54,9 @@ function! s:NERDTree.Close()
         call nerdtree#exec(s:NERDTree.GetWinNum() . " wincmd w", 1)
         close
         if l:useWinId
-            call nerdtree#exec("call win_gotoid(" . l:activeBufOrWin . ")")
+            call nerdtree#exec("call win_gotoid(" . l:activeBufOrWin . ")",0)
         else
-            call nerdtree#exec(bufwinnr(l:activeBufOrWin) . " wincmd w")
+            call nerdtree#exec(bufwinnr(l:activeBufOrWin) . " wincmd w",0)
         endif
     else
         close
--- a/lib/nerdtree/nerdtree.vim
+++ b/lib/nerdtree/nerdtree.vim
@@ -54,9 +54,9 @@ function! s:NERDTree.Close()
         call nerdtree#exec(s:NERDTree.GetWinNum() . " wincmd w", 1)
         call nerdtree#exec("close", 1)
         if l:useWinId
-            call nerdtree#exec("call win_gotoid(" . l:activeBufOrWin . ")",0)
+            call nerdtree#exec("call win_gotoid(" . l:activeBufOrWin . ")", 0)
         else
-            call nerdtree#exec(bufwinnr(l:activeBufOrWin) . " wincmd w",0)
+            call nerdtree#exec(bufwinnr(l:activeBufOrWin) . " wincmd w", 0)
         endif
     else
         close
--- a/lib/nerdtree/opener.vim
+++ b/lib/nerdtree/opener.vim
@@ -220,7 +220,7 @@ function! s:Opener._newVSplit()
     endif
 
     call nerdtree#exec('wincmd p', 1)
-    vnew
+    call nerdtree#exec('vnew', 1)
 
     let l:currentWindowNumber = winnr()
 
--- a/lib/nerdtree/opener.vim
+++ b/lib/nerdtree/opener.vim
@@ -203,7 +203,7 @@ function! s:Opener._newSplit()
         let size = exists("b:NERDTreeOldWindowSize") ? b:NERDTreeOldWindowSize : g:NERDTreeWinSize
         call nerdtree#exec(there, 1)
         exec("silent ". splitMode ." resize ". size)
-        call nerdtree#exec('wincmd p',0)
+        call nerdtree#exec('wincmd p', 0)
     endif
 
     " Restore splitmode settings
@@ -228,7 +228,7 @@ function! s:Opener._newVSplit()
     call g:NERDTree.CursorToTreeWin()
     execute 'silent vertical resize ' . l:winwidth
 
-    call nerdtree#exec(l:currentWindowNumber . 'wincmd w',0)
+    call nerdtree#exec(l:currentWindowNumber . 'wincmd w', 0)
 endfunction
 
 " FUNCTION: Opener.open(target) {{{1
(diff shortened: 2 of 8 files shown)


PhilRunninger:
Greg @wincent ,

If you have some time and you don't mind, I'd like you to take a look at this PR. I remember your comments in #684, and wondered if you would have an opinion on the approach taken here to solve the ping-ponging problem.

If you don't have the time, no worries.

Thanks,
Phil

lib/nerdtree/nerdtree.vim
```diff
@@ -44,19 +44,19 @@ function! s:NERDTree.Close()
...
-        call nerdtree#exec(s:NERDTree.GetWinNum() . " wincmd w")
-        close
+        call nerdtree#exec(s:NERDTree.GetWinNum() . " wincmd w", 1)
+        call nerdtree#exec("close", 1)
         if l:useWinId
-            call nerdtree#exec("call win_gotoid(" . l:activeBufOrWin . ")")
+            call nerdtree#exec("call win_gotoid(" . l:activeBufOrWin . ")",0)
         else
-            call nerdtree#exec(bufwinnr(l:activeBufOrWin) . " wincmd w")
+            call nerdtree#exec(bufwinnr(l:activeBufOrWin) . " wincmd w",0)
```
wincent: Looks like you need a space after the `,` here and two lines above.

lib/nerdtree/opener.vim
```diff
@@ -201,9 +201,9 @@ function! s:Opener._newSplit()
     "resize the tree window if no other window was open before
     if onlyOneWin
         let size = exists("b:NERDTreeOldWindowSize") ? b:NERDTreeOldWindowSize : g:NERDTreeWinSize
-        call nerdtree#exec(there)
+        call nerdtree#exec(there, 1)
         exec("silent ". splitMode ." resize ". size)
-        call nerdtree#exec('wincmd p')
+        call nerdtree#exec('wincmd p',0)
```
wincent: Space after `,`.

lib/nerdtree/opener.vim
```diff
@@ -219,16 +219,16 @@ function! s:Opener._newVSplit()
...
+    call nerdtree#exec('vnew', 1)
 
     let l:currentWindowNumber = winnr()
 
     " Restore the NERDTree to its original width.
     call g:NERDTree.CursorToTreeWin()
     execute 'silent vertical resize ' . l:winwidth
 
-    call nerdtree#exec(l:currentWindowNumber . 'wincmd w')
+    call nerdtree#exec(l:currentWindowNumber . 'wincmd w',0)
```
wincent: And here.

lib/nerdtree/opener.vim
```diff
@@ -321,7 +321,7 @@ function! s:Opener._reuseWindow()
     "check the current tab for the window
     let winnr = bufwinnr('^' . self._path.str() . '$')
     if winnr != -1
-        call nerdtree#exec(winnr . "wincmd w")
+        call nerdtree#exec(winnr . "wincmd w",0)
```
wincent: And here.

lib/nerdtree/opener.vim
```diff
@@ -334,9 +334,9 @@ function! s:Opener._reuseWindow()
     let tabnr = self._path.tabnr()
     if tabnr
         call self._checkToCloseTree(1)
-        call nerdtree#exec(tabnr . 'tabnext')
+        call nerdtree#exec(tabnr . 'tabnext', 1)
         let winnr = bufwinnr('^' . self._path.str() . '$')
-        call nerdtree#exec(winnr . "wincmd w")
+        call nerdtree#exec(winnr . "wincmd w",0)
```
wincent: And here.

lib/nerdtree/ui.vim
```diff
@@ -506,10 +506,10 @@ endfunction
 function! s:UI.toggleZoom()
     if exists("b:NERDTreeZoomed") && b:NERDTreeZoomed
         let size = exists("b:NERDTreeOldWindowSize") ? b:NERDTreeOldWindowSize : g:NERDTreeWinSize
-        exec "silent vertical resize ". size
+        call nerdtree#exec("silent vertical resize ". size,1)
         let b:NERDTreeZoomed = 0
     else
-        exec "vertical resize ". get(g:, 'NERDTreeWinSizeMax', '')
+        call nerdtree#exec("vertical resize ". get(g:, 'NERDTreeWinSizeMax', ''),1)
```
wincent: And on these two.

nerdtree_plugin/fs_menu.vim
```diff
@@ -114,14 +114,14 @@ function! s:promptToDelBuffer(bufnum, msg)
...
-            exec "tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':bnext! ' | endif"
+            call nerdtree#exec("tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':bnext! ' | endif",1)
         else
-            exec "tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':enew! ' | endif"
+            call nerdtree#exec("tabdo windo if winbufnr(0) == " . a:bufnum . " | exec ':enew! ' | endif",1)
         endif
-        exec "tabnext " . s:originalTabNumber
-        exec s:originalWindowNumber . "wincmd w"
+        call nerdtree#exec("tabnext " . s:originalTabNumber,1)
+        call nerdtree#exec(s:originalWindowNumber . "wincmd w",1)
```
wincent: More missing spaces.

nerdtree_plugin/fs_menu.vim
```diff
@@ -141,17 +141,17 @@ function! s:renameBuffer(bufNum, newNodeName, isDirectory)
         let editStr = g:NERDTreePath.New(a:newNodeName).str({'format': 'Edit'})
     endif
     " 1. ensure that a new buffer is loaded
-    exec "badd " . quotedFileName
+    call nerdtree#exec("badd " . quotedFileName,1)
```
wincent: And here.

PR #941, opened 2019-01-28
add option to spawn split from previous window

dragonxlwang:
instead of changing current behavior, add an option to make it flexible

reference: #313

PR #1035, opened 2019-09-09
Make the behavior of window splits consistent

### LINKED ISSUE #327: Opening files by default to the right split?
When I go to NERDTree and press 's', a split happens on the RIGHT side of the screen (I have splitright enabled)

Now if I go back to the NERDTree panel and press 'enter' on another file, the LEFT split gets replaced.

I would like all opening of files to happen on the right split. Is there a way to do this?

PhilRunninger:
### Description of Changes
Closes #941  
This PR addresses changes requested for #941, and expands on the work started there.

Closes #327 
This PR forces NERDTree to respect the user's values for the `splitright` and `splitbelow` settings, and the `g:NERDTreeWinPos` variable. This changes the current behavior of the splits opened by NERDTree, so this will be a new MAJOR release - 6.0.0. The changes are as follows:

1. The window to be split will **always** be the previous window. This could be the one where `:NERDTree` (or `:NERDTreeFocus` or `:NERDTreeToggle`) was invoked. Or maybe a window command (such as `10<C-W>h`) was used to jump directly to the NERDTree window. When in doubt, you can use the `<C-W>p` command to jump between the previous and current windows.
1. The `g:NERDTreeWinPos` variable will **always** be respected. This pertains to a NERDTree created when `g:NERDTreeHijackNetrw` is 1, and Vim is invoked with a directory, as in `vim .`. When splitting the window multiple times in this scenario, NERDTree will stay full-height on the desired side of the screen.
1. The vim setting `splitright` will be be respected. When performing a vertical split (`s` or `gs`), the previous window will be split to the right (`splitright`) or left (`nosplitright`) as directed by the user's setting.
1. The vim setting `splitbelow` will be be respected. When performing a horizontal split (`i` or `gi`), the previous window will be split above (`nosplitbelow`) or below (`splitbelow`) as directed by the user's setting.

These changes bring consistency to the NERDTree split commands, which previously had worked differently for vertical vs horizontal splitting.

---
### New Version Info

- [x] Derive a new version number. Increment the:
    - [x] `MAJOR` version when you make incompatible API changes
    - [ ] `MINOR` version when you add functionality in a backwards-compatible manner
    - [ ] `PATCH` version when you make backwards-compatible bug fixes
- [x] Update [CHANGELOG.md](https://github.com/scrooloose/nerdtree/blob/master/CHANGELOG.md), following the established pattern.
- [ ] Tag the merge commit, e.g. `git tag -a 3.1.4 -m "v3.1.4" && git push origin --tags`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Change Log
 
+#### 6.0...
+- **.0**: Make the behavior of window splits consistent [#1035](https://github.com/scrooloose/nerdtree/pull/1035)
 #### 5.3...
 - **.1**: Fix the `e` key mapping to use netrw if desired [#1031](https://github.com/scrooloose/nerdtree/pull/1031)
 - **.0**: Add file extension and size to sorting capabilities [#1029](https://github.com/scrooloose/nerdtree/pull/1029)
--- a/autoload/nerdtree.vim
+++ b/autoload/nerdtree.vim
@@ -4,6 +4,11 @@ endif
 let g:loaded_nerdtree_autoload = 1
 
 let s:rootNERDTreePath = resolve(expand("<sfile>:p:h:h"))
+
+"FUNCTION: nerdtree#version(...) {{{1
+"  If any value is given as an argument, the entire line of text from the
+"  change log is shown for the current version; otherwise, only the version
+"  number is shown.
 function! nerdtree#version(...)
     let l:changelog = readfile(join([s:rootNERDTreePath, "CHANGELOG.md"], nerdtree#slash()))
     let l:text = 'Unknown'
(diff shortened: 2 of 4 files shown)


PhilRunninger:
@lifecrisis , would you have some time to look at this? It's not a huge change to the codebase, but it is a somewhat radical departure in terms of functionality. To test it, open vim both ways: both with a directory name argument, and without.

1. Once inside NERDTree, open several files, splitting both vertically and horizontally.
1. Go to a specific window you want to split, then open or focus NERDTree, and split-open a file.
1. In each case, see if the right window was split correctly.

There's not a big rush on this; I'll be exercising this branch over the coming days to see if anything shakes loose. I just would like another pair of eyes on it.

PR #1049, opened 2019-10-13
Add VIM built-in package management to read me file

pesarkhobeee:
I am a new user in vim then I found this awesome plugin
The only problem was about installing it inside the new version of vim which I knew we don't have to use an external package manager, therefore, I just added the missing part in order to make it easier to install for the people like myself :)

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -17,6 +17,13 @@ included documentation.
 Installation
 ------------
 
+If you are using VIM version 8 or higher you can use its built-in package management; see `:help packages` for more information. Just run these commands in your terminal:
+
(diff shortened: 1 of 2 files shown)


README.markdown
```diff
@@ -17,6 +17,13 @@ included documentation.
 Installation
 ------------
 
+If you are using VIM version 8 or higher you can use its built-in package management, just run below commands in your terminal:
```
PhilRunninger: Let the reader know where to find more information about packages.

```suggestion
If you are using VIM version 8 or higher you can use its built-in package management; see `:help packages` for more information. Just run these commands in your terminal:
```

pesarkhobeee: I think now it is ready to merge :)

PR #1056, opened 2019-11-13
Menu option, 'copy path to clipboard' is aware of VIM clipboard option

jhzn:
VIM noob here, so code might not be optimal.

My setup is as follows.
NVIM v0.5.0-95-g2e14dffbb, Linux Mint 19.2.

I have this in init.vim
```vim
set clipboard=unnamedplus
```
This enables me to share clipboard between VIM and X clipboard.

The problem is that the menu option in NERDTree copies the file path to the "* register.
This means I can't access the value in the X clipboard.


PhilRunninger:
Looks good. Thanks.

PR #1071, opened 2019-12-31
Lint vimscript, fix errors and warnings, add CI job to review PRs

alerque:
Linting is not as good as full on integration testing, but this is a start. The [vint linter](https://github.com/Kuniwak/vint) for VimL is actually pretty good. It's noisy the first time you run it because it's picking about idiosyncrasies of VIM's language and wants all code to avoid even potential breakage, but it is all good practice and once everything is fixed it's pretty easy to use. Once installed it can be run locally using `vint .` from the project root.

The CI job included here will run the linter in Github Actions and a bot will leave comments on PRs on the line of code where any lint issues are found.

There are a couple flat out bugs that it caught, but most of the changes here were just fixing warnings. I highly recommend this PR ***not* be squashed**! This touches a TON of code, but each commit is split out by the kind of lint issue being fixed and it will be much easier to understand the `git blame` output and potentially `git bisect` if a problem was introduced if they are **not** lumped together.

Please do make sure the plugin still runs as expected. I don't usually use it myself so I'm not a good one to put it through even quick paces.

If you poor through the diff it's probably a good idea to do so a commit at a time so it's not just a mess of changes ;-)

After it is merged I would recommend setting `master` up as a protected branch and requiring the CI job to pass on all PRs before merger. I can set that up in the Github settings is you approve (but I wont' mess with it unless you give the word).

### DIFF
--- a/.github/workflows/vint.yml
+++ b/.github/workflows/vint.yml
@@ -1,0 +1,15 @@
+name: Vint
+on: [push, pull_request]
+jobs:
+  vint:
+    strategy:
+      fail-fast: false
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@master
+    - name: Run vint with reviewdog
+      uses: reviewdog/action-vint@v1.0.1
+      with:
+        github_token: ${{ secrets.github_token }}
+        reporter: github-pr-review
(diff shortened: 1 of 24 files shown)


PhilRunninger:
@alerque , I like the linting that you've added. That will be very handy going forward.

PR #1072, opened 2019-12-31
Update links to repository with new org namespace

alerque:
This should be a no-op as far as the plugin is concerned. I don't know what your changelog protocol is for things like this so I didn't enter one. Let me know if I should.

These links are not broken yet, but someday they might and particularly for people cloning the repository or adding it to their plugin managers it would be nice if they got the canonical location in the first go.


lib/nerdtree/tree_dir_node.vim
```diff
@@ -256,7 +256,7 @@ endfunction
 
 " FUNCTION: TreeDirNode.getDirChildren() {{{1
 " Return a list of all child nodes from "self.children" that are of type
-" TreeDirNode. This function supports http://github.com/scrooloose/nerdtree-project-plugin.git.
+" TreeDirNode. This function supports http://github.com/preservim/nerdtree-project-plugin.git.
```
PhilRunninger: Revert this change. The referenced project has not moved.

alerque: Ooops! Will do.

alerque: Done, good catch. That also happens to fix the merge conflict that was showing up as a result of the massive code style refactoring that just landed.

alerque:
I didn't force push to the master branch, only to this feature branch, so it wouldn't mess up anybody that wasn't actively testing _this_ feature branch.

The rebase (and force push) push was to avoid 3 noisy hard to review commits for one simple action. Not only was there a mistake in the previous one, there was going to be a merge conflict with master. Fixing those was going to end up creating 3 or 4 commits in history just to update these URL. The fix I did kept it so that this is a one-commit PR with a commit that is ready to cleanly merge to (or squash to or rebase onto) master. The result for other devs will be a simple update with no clutter obscuring what was changed or why.

You shouldn't have done a merge when you pulled —and please don't merge the result of _that_ to master. What you could/should do with feature branches like this is rebase when updating them with `git pull --rebase`. At this point if you have a different local history than the remote you can also just `git reset --force` to the remote branch.

PhilRunninger:
Good to know. There is always more git stuff to learn.

PR #1081, opened 2020-01-19
[BUGFIX] Shows error for empty filenames (vim-startify)

### LINKED ISSUE #1059: NERDTreeFind fails with vim-startify or empty buffer
#### Description

`:NERDTreeFind` prints this error when used with `vim-startify`. My guess is this is related.

image

Possible related, because it's working fine with not exisiting/written files:
https://github.com/scrooloose/nerdtree/pull/1043

andys8:
### Description of Changes

There was a fix in #1043 which improves the behavior for non saved filenames. Those are not on disk, but also not empty. This lead to an issue #1059 where actual "empty files" like created with vim-startify or stdin lead to an error. This change fixes this by adapting the order of the tests. It'll print "no file for the current buffer" for empty files.

Closes #1059

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
         - **.PATCH**: Pull Request Title (PR Author) [PR Number](Link to PR)
 -->
 #### 6.4
+- **.6:** NERDTreeFind shows expected message if file doesn't exist e.g. with vim-startify (andys8). [#1081](https://github.com/preservim/nerdtree/pull/1081)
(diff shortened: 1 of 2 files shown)


PhilRunninger:
@andys8 Thanks for finding this and writing a fix for it. I won't have a chance to review it until Saturday. I'm the meantime, please update `CHANGELOG.md` with the new version number. This will be a patch number increment: `6.4.6` Thanks.

andreas-ebot7:
Thanks, no hurry. Updated the changelog :)

PR #1172, opened 2020-09-20
Added table of contents to README.md

staswiner:
Added table of contents to README.md using gh-md-toc


PhilRunninger:
@staswiner While I appreciate the effort you put into this change, I would like to take the README in a different direction. It mentions the NERDTree API, so I'd like to highlight the known plugins that extend NERDTree's functionality: `vim-devicons`, `nerdtree-git-plugin`, etc. I also would like to move all the FAQ material strictly to the Wiki, and better organize that. The Wiki can be edited by anyone, and doesn't require pull requests to be updated.

staswiner:
Thanks for the reply! You’re the boss on this one :)

PR #1200, opened 2021-01-05
Respect NERDTreeCustomOpenArgs when opening bookmark

### LINKED ISSUE #1201: Respect NERDTreeCustomOpenArgs when opening bookmark
#### Description
I have 
```vim
let g:NERDTreeCustomOpenArgs = {'file': {'reuse': 'all', 'where': 't'}, 'dir': {}}
```
but this setting is ignored when opening any new bookmark.

przepompownia:
Closes #1201

### DIFF
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -130,6 +130,8 @@ function! s:initCustomOpenArgs() abort
         endfor
     endfor
 
+    let g:NERDTreeCustomOpenArgs = l:defaultOpenArgs
+
     return extend(l:customOpenArgs, l:defaultOpenArgs, 'keep')
 endfunction
 
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -500,9 +500,15 @@ function! nerdtree#ui_glue#openBookmark(name) abort
     endtry
     if l:bookmark.path.isDirectory
         call l:bookmark.open(b:NERDTree)
-    else
-        call l:bookmark.open(b:NERDTree, {'where': 'p'})
+        return
     endif
+
+    let l:openArgs = get(g:, 'NERDTreeCustomOpenArgs', {})
+    let l:options = (
+                \ type(l:openArgs) ==# v:t_dict
+                \ && has_key(l:openArgs, 'file')
+                \ ) ? l:openArgs.file : {'where': 'p'}
+    call l:bookmark.open(b:NERDTree, l:options)
 endfunction
 
 " FUNCTION: s:openHSplit(target) {{{1
--- a/autoload/nerdtree/ui_glue.vim
+++ b/autoload/nerdtree/ui_glue.vim
@@ -124,9 +124,10 @@ function! s:initCustomOpenArgs() abort
         endif
 
         for l:optionName in keys(l:defaultOpenArgs[l:typeKey])
-            if !s:validateType(get(l:customOpenArgs[l:typeKey], l:optionName, v:null), v:t_string)
-                let l:customOpenArgs[l:typeKey][l:optionName] = l:defaultOpenArgs[l:typeKey][l:optionName]
+            if s:validateType(get(l:customOpenArgs[l:typeKey], l:optionName, v:null), v:t_string)
+                continue
             endif
+            let l:customOpenArgs[l:typeKey][l:optionName] = l:defaultOpenArgs[l:typeKey][l:optionName]
         endfor
     endfor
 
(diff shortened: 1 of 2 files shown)


PhilRunninger:
The function you changed in this pull request is the one used by the `t` key for bookmarks. By design, this does not use the `g:NERDTreeCustomOpenArgs`. The **Custom Open** function is linked to the `Enter` key, and is intended to provide a second, customizable way to open files. Your setting will work as intended if you use `<CR>` instead of `t`.

przepompownia:
- `Enter` on any bookmark item works as intended
- `OpenBookmark` still opens buffer in the current window.

PhilRunninger:
Oh. My mistake. This applies to the `:OpenBookmark` command. I'll have to think about this one. This is going to change the way the command works in some circumstances, but I'm not sure it will make much difference. This seems like a command that doesn't get used very much.

autoload/nerdtree/ui_glue.vim
```diff
@@ -500,9 +500,15 @@ function! nerdtree#ui_glue#openBookmark(name) abort
...
-        call l:bookmark.open(b:NERDTree, {'where': 'p'})
+        return
     endif
+
+    let l:openArgs = get(g:, 'NERDTreeCustomOpenArgs', {})
+    let l:options = (
+                \ type(l:openArgs) ==# v:t_dict
+                \ && has_key(l:openArgs, 'file')
+                \ ) ? l:openArgs.file : {'where': 'p'}
+    call l:bookmark.open(b:NERDTree, l:options)
```
PhilRunninger, line 511:
```suggestion
     else
        call l:bookmark.open(b:NERDTree, s:initCustomOpenArgs().file)
    endif
```

There is a simpler way to write this, by using the function set up to initialize the CustomOpenArgs. If you want to move the type validation to that function, I'd be fine with that.

przepompownia: How it works now:
```vim
function! s:initCustomOpenArgs(customOpenArgs) abort
    let l:defaultOpenArgs = {'file': {'reuse': 'all', 'where': 'p'}, 'dir': {}}
    let l:customOpenArgs = a:customOpenArgs

    if !s:validateType(l:customOpenArgs, v:t_dict)
        return l:defaultOpenArgs
    endif

    for l:typeKey in keys(l:defaultOpenArgs)
        if !s:validateType(get(l:customOpenArgs, l:typeKey, {}), v:t_dict)
            let l:customOpenArgs[l:typeKey] = l:defaultOpenArgs[l:typeKey]
            continue
        endif

        for l:optionName in keys(l:defaultOpenArgs[l:typeKey])
            if !s:validateType(get(l:customOpenArgs[l:typeKey], l:optionName, v:null), v:t_string)
              let l:customOpenArgs[l:typeKey][l:optionName] = l:defaultOpenArgs[l:typeKey][l:optionName]
            endif
        endfor
    endfor

    return extend(l:customOpenArgs, l:defaultOpenArgs, 'keep')
endfunction

function! s:validateType(variable, type)
    if type(a:variable) is# a:type
        return v:true
    endif

    return v:false
endfunction

function! s:testInitCustomOpenArgs(options, expectedResult)
    call assert_equal(a:expectedResult, s:initCustomOpenArgs(a:options))
endfunction

function! s:testValidateType(variable, expectedResult)
  call assert_true(s:validateType(a:variable, a:expectedResult))
endfunction

let v:errors = []
let defaultOpenArgs = {'file': {'reuse': 'all', 'where': 'p'}, 'dir': {}}

call s:testValidateType({}, v:t_dict)
call s:testValidateType(defaultOpenArgs['file']['where'], v:t_string)

call s:testInitCustomOpenArgs(1, defaultOpenArgs)
call s:testInitCustomOpenArgs({'file': 1}, defaultOpenArgs)
call s:testInitCustomOpenArgs({'file': {}, 'dir': 'foo'}, defaultOpenArgs)
call s:testInitCustomOpenArgs(
      \ {'file': {'reuse': [1], 'where': 'w', 'foo': []}},
      \ {'file': {'reuse': 'all', 'where': 'w', 'foo': []}, 'dir': {}}
      \ )
call s:testInitCustomOpenArgs(
      \ {'file': {'reuse': [1], 'foo': []}},
      \ {'file': {'reuse': 'all', 'where': 'w', 'foo': []}, 'dir': {}}
      \ )

if !len(v:errors)
  echom 'Success!'
  finish
endif

echoerr v:errors
```

autoload/nerdtree/ui_glue.vim
```diff
@@ -500,9 +530,10 @@ function! nerdtree#ui_glue#openBookmark(name) abort
     endtry
     if l:bookmark.path.isDirectory
         call l:bookmark.open(b:NERDTree)
-    else
-        call l:bookmark.open(b:NERDTree, {'where': 'p'})
+        return
```
przepompownia, line 533: I left quick returning to avoid unnecessary nesting level.

PhilRunninger (approved):

PR #1219, opened 2021-02-09
Removed directory separator from sort key

### LINKED ISSUE #1218: Edge case in comparisons for directories with similar prefixes?
#### Self-Diagnosis

- [x] I have searched the [issues](https://github.com/scrooloose/nerdtree/issues) for an answer to my question.
- [x] I have reviewed the NERDTree documentation. `:h NERDTree`
- [x] I have reviewed the [Wiki](https://github.com/scrooloose/nerdtree/wiki).
- [x] I have searched the web for an answer to my question.

#### State Your Question

When comparing two directories that have similar prefixes (i.e. "project" and "project-subproject") the actual comparison happening behind the scenes is more accurately represented as "project/" vs. "project-" because the paths include the separator (as intended when being rendered).  This actually makes "project" which would ordinarily appear at the beginning of the of the grouped directories, appear at the very end of the group.

Quite specifically, this can occur in Rust development when many sub-projects are present; it can be odd having to scroll to the bottom for what would be your core crate.

```
project-rs/
    project-subproject1/
    ...
    project-subproject5/
    project/
```

~Anyway, upon some digging I found this peculiar function which is responsible for handling comparisons between nodes:~

~Diving in will show some strange behavior that I don't think is intentional; both `sortKey1` and `sortKey2` appear to be pairs, the first value being a number indicating whether the node is a directory or a file and the second value being the actual path name (which happens to include the path separator).~

~Knowing this, it seems as if the `min(...)` comparison was actually intended for the path name lengths but now just returns 2 and always compares the types of the pair's items?~

Edit: After informing myself a little on how vim handles comparisons, it seems as though that part of the node comparison logic works as intended!  I am very sorry for my mistake but still I'm curious if it's intentional to include the directory separator in the comparison?

gladiator:
### Description of Changes
Closes #1218 

Directories had an additional separator appended to them which caused improper comparisons for other directories that shared similar prefixes.

---
#### Author's Instructions
- [x] Derive a new `MAJOR.MINOR.PATCH` version number. Increment the:
- [x] Update [CHANGELOG.md](https://github.com/scrooloose/nerdtree/blob/master/CHANGELOG.md), following the established pattern.
    ```bash
    git checkout master
    git pull
    ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
         - **.PATCH**: Pull Request Title (PR Author) [PR Number](Link to PR)
 -->
 #### 6.10
+- **.4**: Removed directory separator from sort key (Daniel E) [#1219](https://github.com/preservim/nerdtree/pull/1219)
 - **.3**: Add new FAQ and answer: How to prevent buffers replacing NERDTree. (PhilRunninger) [#1215](https://github.com/preservim/nerdtree/pull/1215)
 - **.2**: New menu command: Run a system command in this directory. (PhilRunninger) [#1214](https://github.com/preservim/nerdtree/pull/1214)
 - **.1**: Escape quotation marks so they can be used in key mappings. (PhilRunninger) [#1213](https://github.com/preservim/nerdtree/pull/1213)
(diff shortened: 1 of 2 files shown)


PhilRunninger (approved):
Good catch! Thanks for the fix. 👍

PR #1274, opened 2021-10-08
On load show which files were not loaded

rmonico:
Just a small patch to show which files couldn't be read.


alerque:
Thanks for this. Since we didn't hear back on the minor feedback above another contributor jumped in with that little fix. The commit from this PR was included in the later one, so this is merged and thanks for the contribution!

PR #1348, opened 2023-02-05
New menu command: Change selected node permissions

mjkloeckner:
### Description of Changes

Adds a new option in the menu to change the selected node permissions. When invoking this menu option - m, C - it prompts the user for a new mode (i.e.: `+x`) and then it uses that input for executing `chown` in the selected node.

Does not work in Windows for now.

---
#### Author's Instructions
    ```bash
    git checkout master
    git pull
    ```

### DIFF
--- a/nerdtree_plugin/fs_menu.vim
+++ b/nerdtree_plugin/fs_menu.vim
@@ -45,6 +45,7 @@ call NERDTreeAddMenuItem({'text': (has('clipboard')?'copy (p)ath to clipboard':'
 
 if has('unix') || has('osx')
     call NERDTreeAddMenuItem({'text': '(l)ist the current node', 'shortcut': 'l', 'callback': 'NERDTreeListNode'})
+    call NERDTreeAddMenuItem({'text': '(C)hange node permissions', 'shortcut':'C', 'callback': 'NERDTreeChangePermissions'})
 else
     call NERDTreeAddMenuItem({'text': '(l)ist the current node', 'shortcut': 'l', 'callback': 'NERDTreeListNodeWin32'})
 endif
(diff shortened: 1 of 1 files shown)


kaboussi (approved):

nerdtree_plugin/fs_menu.vim
```diff
@@ -333,6 +334,29 @@ function! NERDTreeListNodeWin32()
     call nerdtree#echo('node not recognized')
 endfunction
 
+" FUNCTION: NERDTreeChangePermissions() {{{1
+function! NERDTreeChangePermissions()
+    let l:node = g:NERDTreeFileNode.GetSelected()
+    let l:prompt = "change node permissions: "
```
rzvxa, line 340: @mjkloeckner what do you think about adding an example to the prompt so the user knows it's chmod?
`let l:prompt = "change node permissions(ex: a+x): "`
Or maybe writing chmod in the prompt to suggest that it uses chmod underneath?
`let l:prompt = "change node permissions(chmod): "`
Other than that I think it's perfect and I personally would love to use it

alerque: Will this work on Windows at all?

rzvxa: No, This option is only showing up on UNIX systems, Windows permission scheme is very different from UNIX so it needs a completely different implementation for Windows platforms(windows uses access control list which more optimized towards GUI usage)

rzvxa: But we already have a few platform-specific options for Linux and Mac which is missing on Windows, For example, `NERDTreeRevealInFinder` and `NERDTreeRevealFileLinux` are implemented for mac and linux but there is no option implemented for revealing in windows explorer

alerque: Yes I missed that this was gated behind OS detection. In that case yes I second some clarification that this is a string that will be passed to `chmod`.

Also is there somewhere the key binding should be documented?

rzvxa: > Also is there somewhere the key binding should be documented?

Good point, I can't find any documentation for fs_menu operations in the document.
There is a dedicated section for `NERDTreeMappings` which covers normal keybindings, but no fs-specific operation is covered there, There is a section for `NERDTreeMenu` which covers their operations on a very high-level and abstract basis(Explains menu system and NERDTreeMenuAPI)

rzvxa: Maybe we should add some documentation for these options in fs_menu

alerque: Adding docs does seem to be in order, but since the related functions are also undocumented it doesn't have to hold up this PR. If we do go ahead with this PR with no documentation we should at least open an issue noting the lack of documentation for fs_menu features.

alerque (approved):

nerdtree_plugin/fs_menu.vim
rzvxa: I'm going to add chmod to the end of the prompt message to make sure the user knows it does a chmod with the given parameters, If anyone has a better idea for this don't hesitate to inform me about it.
By definition it does call shell commands and can be used to execute malicious code but I don't think it's a problem for an editor plugin. A bar symbol or even a semi-colon can be used to terminate chmod and run some other command.

PR #1359, opened 2023-07-04
Fix [...] directories not showing children

kai-patel:
### Description of Changes

Correctly escape directories named `[<dir_name>]` on Windows. Per `:help wildcards`:

```
To avoid the special meaning of the wildcards prepend a backslash.  However,
on MS-Windows the backslash is a path separator and "path\[abc]" is still seen
as a wildcard when "[" is in the 'isfname' option.  A simple way to avoid this
is to use "path\[[]abc]", this matches the file "path\[abc]".
```

This fix simply implements this pattern for `TreeDirNode` `l:pathSpec`, and only on Windows - as I haven't seen any documentation showing this bug on other systems.

Before this change, children of directories with square-bracketed names would not be shown (see issues below for examples).

This is related to, but does not close necessarily vim/vim#12629, neovim/neovim#24233

---
#### Author's Instructions
    ```bash
    git checkout master
    git pull
    ```

### DIFF
--- a/lib/nerdtree/tree_dir_node.vim
+++ b/lib/nerdtree/tree_dir_node.vim
@@ -278,6 +278,10 @@ function! s:TreeDirNode._glob(pattern, all)
     else
         let l:pathSpec = escape(fnamemodify(self.path.str({'format': 'Glob'}), ':.'), ',')
 
+        if nerdtree#runningWindows()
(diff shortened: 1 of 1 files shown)


Gustavo-DMS (approved):

rzvxa (approved):
Great work thank you, I've tested it on Windows, Works beautifully.

PR #1375, opened 2023-10-17
Case Sensitive Move Operation

### LINKED ISSUE #1373: `(m)ove the current node` shortcut doesn't treat filepaths as case sensitive
#### Description
- The  `(m)ove the current node` menu shortcut doesn't treat file paths as case sensitive.  This prevents the command from detecting changes to the capitalization of letters in a move command (ex: it treats `~/project/Test.go` and `~/project/test.go` as equivalent).

#### Environment
- Operating System: `macOS Sonoma (14.0)`
- Vim/Neovim version `:version`: vim `9.0.1544`
- NERDTree version, found on first line of quickhelp `?`: `6.10.16`
- Are you using any of these NERDTree-dependent plugins?  
    - [ ] [Xuyuanp/nerdtree-git-plugin](https://github.com/Xuyuanp/nerdtree-git-plugin)
    - [ ] [ryanoasis/vim-devicons](https://github.com/ryanoasis/vim-devicons)
    - [ ] [tiagofumo/vim-nerdtree-syntax-highlight](https://github.com/tiagofumo/vim-nerdtree-syntax-highlight)
    - [ ] [scrooloose/nerdtree-project-plugin](https://github.com/scrooloose/nerdtree-project-plugin)
    - [ ] [PhilRunninger/nerdtree-buffer-ops](https://github.com/PhilRunninger/nerdtree-buffer-ops)
    - [ ] [PhilRunninger/nerdtree-visual-selection](https://github.com/PhilRunninger/nerdtree-visual-selection)
    - [ ] [jistr/vim-nerdtree-tabs](https://github.com/jistr/vim-nerdtree-tabs)
    - [ ] Others (specify):
- Provide a minimal **.vimrc** file that will reproduce the issue.
    - No custom run commands needed

#### Steps to Reproduce the Issue
1. Create a file named `test.txt` 
2. Open said file in `vim` & open `NERDTREE`
3. Move the file to the same path using the `mm` command and name it `Test.txt`

#### Current Behavior (Include screenshots where appropriate.)
-  The following output is shown: `NERDTREE: This destination already exists. Try again`

#### Expected Result
- Moves the file as expected

rzvxa:
### Description of Changes
Closes #1373 and potentially many unresolved issues on the subject of case insensitivity.
In this commit, I've introduced a new option called `NERDTreeCaseInsensitiveFS` which could also be used in the future to resolve similar issues in other places.
Until now the Move Operation used to check if the destination path already exists or not, On case-insensitive file systems this check fails if the user wants to move the file to a location with a different casing (for example moving/renaming `foo.bar` to `FOO.BAR`) as both paths point to the same file. With this new option, we can have an extra check to see whether or not the new destination is the same path in a different casing, and if that's the case we can ignore this false positive check.
With this new change came a wrong artifact, After the rename operation we check if an open buffer to this path exists or not.
Then we will point every window to that buffer to a new buffer for the new file path, vim doesn't acknowledge the new file path if it already has one open for the old casing file, I've used a temp buffer to compensate for this issue. Now we first redirect every window to a temp buffer, close the old buffer, and point windows looking at the temp buffer to this newly created one.
This part is written very clearly and has a lot of comments explaining each step of the process in the code itself.

---
### New Version Info
@alerque Sadly we skipped a few commits versioning, I would like to make a PR adding versions / CHANGELOG to the old merged PRs.
We can either block this PR until we fix the change log, or go ahead and merge it and give it a change log in the fix PR.
#### Author's Instructions
    ```bash
    git checkout master
    git pull
    ```

### DIFF
--- a/doc/NERDTree.txt
+++ b/doc/NERDTree.txt
@@ -673,6 +673,9 @@ the NERDTree. These settings should be set in your vimrc, using `:let`.
 
 |NERDTreeAutoCenterThreshold| Controls the sensitivity of autocentering.
 
+|NERDTreeCaseInsensitiveFS|   Tells the NERDTree whether or not it is
+                            running in on a case sensitive file system.
+
 |NERDTreeCaseSensitiveSort|   Tells the NERDTree whether to be case
                             sensitive or not when sorting nodes.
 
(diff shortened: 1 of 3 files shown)


alerque (approved):

alerque (approved):

rzvxa:
I've gone through some iterations and tried to streamline our approach to potentially fix other similar issues using it.
I can confirm it works as intended on Windows, But it affects Mac users the most as there were some case-insensitive checks in place for Windows but not for macOS, and with this PR I've unified both operating systems under the new setting `g:NERDTreeCaseSensitiveFS`

alerque (approved):

kevinkowalew:
I tested both `0` and `3` on macOS this morning and it achieves the desired behavior, should be good to merge 🎉

alerque:
@kevinkowalew Thanks for testing and reporting, that's super helpful to move things forward!

rzvxa:
Thanks to @kevinkowalew we have confirmation that these changes work on the macOS, I would've waited for @dangibson to give some feedback on this PR because he's also working on a similar issue in #1346 but since these changes are pretty essential and minor implementation details can always be changed in the subsequent patches I'm going to move forward with merging it into the upstream.

PR #1381, opened 2023-10-19
Clarified the NERDTreeChangePermissions prompt

rzvxa:
### Description of Changes
Related to #1348     

---
#### Author's Instructions
    ```bash
    git checkout master
    git pull
    ```

### DIFF
--- a/nerdtree_plugin/fs_menu.vim
+++ b/nerdtree_plugin/fs_menu.vim
@@ -374,7 +374,7 @@ endfunction
 " FUNCTION: NERDTreeChangePermissions() {{{1
 function! NERDTreeChangePermissions()
     let l:node = g:NERDTreeFileNode.GetSelected()
-    let l:prompt = "change node permissions(chmod): "
+    let l:prompt = "change node permissions (chmod args): "
     let l:newNodePerm = input(l:prompt)
 
     if !empty(l:node)
--- a/nerdtree_plugin/fs_menu.vim
+++ b/nerdtree_plugin/fs_menu.vim
@@ -337,7 +337,7 @@ endfunction
 " FUNCTION: NERDTreeChangePermissions() {{{1
 function! NERDTreeChangePermissions()
     let l:node = g:NERDTreeFileNode.GetSelected()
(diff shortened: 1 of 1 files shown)


nerdtree_plugin/fs_menu.vim
```diff
@@ -374,7 +374,7 @@ endfunction
 " FUNCTION: NERDTreeChangePermissions() {{{1
 function! NERDTreeChangePermissions()
     let l:node = g:NERDTreeFileNode.GetSelected()
-    let l:prompt = "change node permissions: "
+    let l:prompt = "change node permissions(chmod): "
```
alerque, line 377:
```suggestion
    let l:prompt = "change node permissions (chmod MODE): "
```

If this needs clarifying, why not be even more explicit like this?

alerque (changes requested):

nerdtree_plugin/fs_menu.vim
rzvxa: I'm all up for making it clearer, But I'm not sure if chomd MODE can give any more information about the command, I thought of chmod word before the prompt as it mimics the basic schematics of chmod call from the shell.
`change node permissions (chmod): a+x` is really similar to the resulting shell command which is `$ chmod a+x path/to/the/file`.
I don't have any strong objection against either of them and now that I have explained my reasoning I would appreciate it if you make the call about it.

alerque: I can see your point, I don't feel super strongly either way. I *do* feel strongly about the need for a space before the opening parenthesis. Fix the space for sure but then pick whatever detail you want. I thought of `(chmod flags)` and `(chmod args)` too. No strong sense of one of them being better than the others. Your call.

rzvxa: I think `chmod args` is a brilliant choice as the user can do something like `u=rw,og=r` as well as `-R +x` so these aren't just sole flags. I'm going to go forward with that one at least for now until someone suggests a better way of prompting this.

alerque (approved):

PR #1237, opened 2021-04-08
This adds split explorer capability to NERDTree

tempoz:
### Description of Changes

This adds a `g:NERDTreeSplexMode` boolean setting that, when activated, causes NERDTree to [act as split explorer-style file browser instead of a project drawer](http://vimcasts.org/blog/2013/01/oil-and-vinegar-split-windows-and-project-drawer/).

### Motivation for Changes

I was looking for a way to browse my filesystem in vim and I came across NERDTree, which I love the look and feel of, but in my search I had also come across many articles extolling the virtues of split explorer-style browsers. Given my restricted monitor space and preference for more intuitive window-splitting, I decided to try to do both. I understand entirely if the project maintainers do not wish to support this feature, but since I'd gone to the trouble of developing it, I thought I'd offer it up.

---
#### Author's Instructions
- [x] Derive a new `MAJOR.MINOR.PATCH` version number. Increment the:
    - `MAJOR` version when you make incompatible API changes
    - `MINOR` version when you add functionality in a backwards-compatible manner
    - `PATCH` version when you make backwards-compatible bug fixes
- [x] Update [CHANGELOG.md](https://github.com/scrooloose/nerdtree/blob/master/CHANGELOG.md), following the established pattern.
#### Collaborator's Instructions
- [ ] Review [CHANGELOG.md](https://github.com/scrooloose/nerdtree/blob/master/CHANGELOG.md), suggesting a different version number if necessary.
- [ ] After merging, tag the commit using these (Mac-compatible) bash commands:
    ```bash
    git checkout master
    git pull
    sed -n "$(grep -n -m2 '####' CHANGELOG.md | cut -f1 -d: | sed 'N;s/\n/,/')p" CHANGELOG.md | sed '$d'
    git tag -a $(read -p "Tag Name: " tag;echo $tag) -m"$(git show --quiet --pretty=%s)";git push origin --tags
    ```

PR #1389, opened 2023-12-12
Add split explorer to NERDTree

msibal6:
### Description of Changes
This adds a new user defined command `NERDTreeExplorer` to a split explorer for a directory. 

### Motivation
I toggle NERDTree often after opening my file, so I tend to already use it more as an explorer than a drawer. I saw [#1237 for adding split explorer](https://github.com/preservim/nerdtree/pull/1237) had a pending action item, so I decided to implement it since it was a simple add when implemented as a command.

I also want to give a quick shout out to @rzvxa for taking the initiative to handling the majority of these most recent PRs since Phil's retirement. It's cool to see a tool live on even when it is very robust and there is not a bunch of changes happening. Thanks to you and everybody for continuing to support this tool and preventing bit rot into it! 😁

### DIFF
--- a/lib/nerdtree/creator.vim
+++ b/lib/nerdtree/creator.vim
@@ -40,6 +40,13 @@ endfunction
 
 " FUNCTION: s:Creator.CreateExplorerTree(dir) {{{1
 function! s:Creator.CreateExplorerTree(dir)
+    try
+        let path = g:NERDTreePath.New(a:dir)
+    catch /^NERDTree.InvalidArgumentsError/
+        call nerdtree#echo('Invalid directory name:' . a:dir)
+        return
+    endtry
+
     let creator = s:Creator.New()
     if getbufinfo('%')[0].changed
         let l:splitLocation = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'top' ? 'topleft ' : 'botright '
--- a/lib/nerdtree/creator.vim
+++ b/lib/nerdtree/creator.vim
@@ -38,8 +38,8 @@ function! s:Creator.BufNamePrefix()
     return 'NERD_tree_'
 endfunction
 
-" FUNCTION: s:Creator.CreateExplorerTree(dir) {{{1
-function! s:Creator.CreateExplorerTree(dir)
+" FUNCTION: s:Creator.CreateExploreTree(dir) {{{1
+function! s:Creator.CreateExploreTree(dir)
     try
         let path = g:NERDTreePath.New(a:dir)
     catch /^NERDTree.InvalidArgumentsError/
@@ -48,12 +48,17 @@ function! s:Creator.CreateExplorerTree(dir)
     endtry
 
     let creator = s:Creator.New()
-    if getbufinfo('%')[0].changed
+    if getbufinfo('%')[0].changed && !&hidden && !&autowriteall
         let l:splitLocation = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'top' ? 'topleft ' : 'botright '
         let l:splitDirection = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'right' ? 'vertical' : ''
+        silent! execute l:splitLocation . l:splitDirection . ' new'
+    else
+        silent! execute 'enew'
     endif
-    silent! execute l:splitLocation . l:splitDirection . ' new'
+
     call creator.createWindowTree(a:dir)
+    "we want windowTree buffer to disappear after moving to any other buffer
+    setlocal bufhidden=wipe
 endfunction
 
 " FUNCTION: s:Creator.CreateTabTree(a:name) {{{1
--- a/lib/nerdtree/creator.vim
+++ b/lib/nerdtree/creator.vim
@@ -38,6 +38,17 @@ function! s:Creator.BufNamePrefix()
     return 'NERD_tree_'
 endfunction
 
+" FUNCTION: s:Creator.CreateExplorerTree(dir) {{{1
+function! s:Creator.CreateExplorerTree(dir)
+    let creator = s:Creator.New()
+    if getbufinfo('%')[0].changed
+        let l:splitLocation = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'top' ? 'topleft ' : 'botright '
+        let l:splitDirection = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'right' ? 'vertical' : ''
+    endif
+    silent! execute l:splitLocation . l:splitDirection . ' new'
+    call creator.createWindowTree(a:dir)
+endfunction
+
 " FUNCTION: s:Creator.CreateTabTree(a:name) {{{1
 function! s:Creator.CreateTabTree(name)
     let creator = s:Creator.New()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,6 @@
             - Pull Request Title n (PR Author) [PR Number](Link to PR)
 -->
 #### 7.0
-- **.2**: Add split explorer capability to NERDTREE (msibal6) [#1389](https://github.com/preservim/nerdtree/pull/1389)
 - **.1**:
     - Fix NERDTreeFind to handle directory case sensitivity. (dangibson) [#1387](https://github.com/preservim/nerdtree/pull/1387)
     - New Show file lines toggle. (hsnks100) [#1384](https://github.com/preservim/nerdtree/pull/1384)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
             - Pull Request Title n (PR Author) [PR Number](Link to PR)
 -->
 #### 7.0
-- **.2**: Add split explorer capability to NERDTREE (msibal6) [#1368](https://github.com/preservim/nerdtree/pull/1368)
+- **.2**: Add split explorer capability to NERDTREE (msibal6) [#1389](https://github.com/preservim/nerdtree/pull/1389)
 - **.1**:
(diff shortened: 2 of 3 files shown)


rzvxa:
Hi,

I appreciate your kind words and more than that I appreciate this PR, it was something I thought of doing in the past few weeks, And I'm so happy to see that somebody else stepped up to implement it.
The moment I saw the title in my email I thought of #1237 which is a useful feature since interacting with file trees in a split fashion helps with the amount of eye movement you need to make to keep track of stuff happening.
As you mentioned it is pretty straightforward to implement it as a command instead of a configuration since the underlying functionality is there and we just need to expose it as a command to the users.

Your PR by itself is completely fine in my opinion but I've found a few quirks in the `CreateWindowTree` function call.
For example, I get weird behavior from it while running in a modified buffer:
WindowsTerminal_YFKmvvj3lY

This function is being used for replacing the current directory buffer which can not be modified so this wasn't an issue until now.
I'm not sure if the author of #1237 has done anything to prevent this issue, But we need to either fix the problem on this PR or block merging this one until we fix the issues in `CreateWindowTree` via another merge request(s).

On the other hand, we are moving away from the previous versioning and going towards automated changelogs via [Conventional Commits](https://www.conventionalcommits.org) and multi-bulletin patches, So there is no need for editing the CHANGELOG and I would be grateful if you revert it. The final merge commit message is going to be used to write the changelogs.

Let me know if you like to work on tracking this and other possible issues with this new use of CreateWindowTree.

PS: In your fashion, I also would like to give a shoutout to @alerque who despite his busy schedule has always spent some time reviewing changes to this project.

msibal6:
Thank you for the quick response! I see now that #1237  used `enew` each time before calling `CreateWindowTree`, so your proposed check is correct. I would like to work on tracking this new usage and any issues for 'CreateWindowTree` if you and @alerque are okay with that. Thank you! 😄

rzvxa:
In my opinion, since the changes are pretty relevant to the feature itself, It is appropriate to include them in this PR.
Just one thing, Be mindful of other parts of the code that call this function in case you want to change the underlying function itself.

msibal6:
Hi @alerque, like @rzvxa already mentioned, thank for your sign off and continued support on this plugin! It is cool to be a small part of the collaboration that can happen on these open source projects. 

@rzvxa, I created a new function to avoid having to mess with the underlying function and fixing any changes that might be needed for other dependent functions. The split on a modified buffer addresses the situation you mentioned and we are still able to leverage the 'CreateWindowTree' function. Please let me know if you have any concerns. Thank you!

lib/nerdtree/creator.vim
```diff
@@ -38,6 +38,24 @@ function! s:Creator.BufNamePrefix()
...
+function! s:Creator.CreateExplorerTree(dir)
+    try
+        let path = g:NERDTreePath.New(a:dir)
+    catch /^NERDTree.InvalidArgumentsError/
+        call nerdtree#echo('Invalid directory name:' . a:dir)
+        return
+    endtry
+
+    let creator = s:Creator.New()
+    if getbufinfo('%')[0].changed
```
rzvxa, line 51: Hi, Thanks for this quick implementation, I like this approach with a new function instead of touching the `createWindowTree` function itself. I have a few thoughts about this section, I appreciate what you have done here with the splitting, But doesn't it make more sense if we replace these lines:
```vim
    if getbufinfo('%')[0].changed
        let l:splitLocation = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'top' ? 'topleft ' : 'botright '
        let l:splitDirection = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'right' ? 'vertical' : ''
    endif
    silent! execute l:splitLocation . l:splitDirection . ' new'
```
with this:
```vim
    if !isdirectory(expand('%:p:r'))
        silent! execute 'enew'
    end
```
Since it makes the behavior more like the netrw Explore command across the board. (also on that note, what do you think about renaming it from `NERDTreeExplorer` to `NERDTreeExplore`?)

msibal6: Hi, the splitting to a new window only occurs when the current buffer is modified. This is to address the behavior on a modified buffer that you mentioned.

rzvxa: Maybe I'm getting it wrong, but we do not discard changes if we just create a new buffer. the modified buffer will continue to exist afterward.
Explore does the same thing, It creates a new netrw buffer and just changes the buffer in the active window(instead of closing or saving it).

We can also introduce splitting behavior with a new command like `NERDTreeSplitExplore` but instead of limiting it to the modified buffers we can apply it to every situation for the sake of consistency; With this change, users can choose between either opening a tree in the current window or create a new split for it without random behavior when the state of file changes.

If you have any strong opposition against doing it this way just let me know so we can talk about it.

msibal6: On a modified buffer, we cannot create a new buffer using `enew`. From documentation, enew "fails when changes
have been made to the current buffer, unless 'hidden
' is set or 'autowriteall' is set and the file can be written." I would say that for most buffers, 'autowriteall' and 'hidden' are not set like in my gif below. After modifying a nerdtree source file, I use `Explore` and it opens a netrw buffer in a new split. This is because a hidden buffer is the only buffer that can have changes without being the active buffer.
output

Because I want to avoid handling or modifying the 'hidden' and 'autowriteall' options or creating separate similar options for nerdtree, I split like `Explore` command on a modified buffer. Then on an unmodified buffer, I replace the current buffer with the nerdtree window tree in the same way as the netrw buffer replacing the current buffer. 

Regarding a `NERDTreeSplitExplore` command, I am open to creating that. However, I want to come to an agreement on what `NERDTreeExplore` should do first. Thanks!

rzvxa: Hey, Sorry for my late response; At the moment I'm a little busy with my day job, and I had to do some testing before I can speak about this PR.
I wasn't aware of this behavior since I always had `set hidden` in my vimrc. Thanks for being so patient with me and informing me about these flags. Now that we are on the same page, wouldn't it be better if we actually check for these flags and act accordingly? After all, we both want to make this command behave similarly to the Explore command.
So here is what I came up with for this part, Give it a look and let me know if there is anything else we should be anticipating in order to keep these 2 commands more aligned.
```vim
    if getbufinfo('%')[0].changed && !&hidden && !&autowriteall
        let l:splitLocation = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'top' ? 'topleft ' : 'botright '
        let l:splitDirection = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'right' ? 'vertical' : ''
        silent! execute l:splitLocation . l:splitDirection . ' new'
    elseif !isdirectory(expand('%:p:r'))
        silent! execute 'enew'
    endif

```

msibal6: No worries on the late reply! I understand if you have to handle your day job first. It is your day job after all 😆. Thank you for your questions! They are forcing me to be more thorough and examine my own assumptions in my own explanations. I am working to be more thorough without being too pedantic.

Coming back to the code, I agree with the handling for `hidden` and `autowriteall` and I will add that in the next commit. For the  `!isdirectory(expand('%:p:r'))` conditional, why do you not want to `enew` on a di?

msibal6: No worries on the late reply! I understand if you are busy with your day job. It is your day job after all 😆. Thank you for your questions. They forced me to be more thorough and examine my own assumptions in my explanations. Please let me know though if I become too thorough and approach being pedantic. 

Back to the code, I agree with your handling of the `hidden` and `autowriteall` flags. I'll push a commit with all the changes we've discussed after one last discussion lol. For the `!isdirectory(expand('%:p:r'))` conditional, why do you not want to use `enew` on a directory buffer? If we are using nerdtree only, when would we be in a directory buffer to not call `enew`? This is for clarification only and I do not see any impact to keeping this aligned to `:Explore`. Thanks!

rzvxa: I was under the impression that createWindowTree will set `bufhidden` to `wipe` and this way we can prevent the NERDTree buffer from appearing in the history when we run it multiple times on the same buffer; But I was wrong about that, Turns out that common buffer options sets the `bufhidden` to `hide` and it can still appear as the previously visited location.

In order to achieve this result I had to change it to this.
```vim
    if getbufinfo('%')[0].changed && !&hidden && !&autowriteall
        let l:splitLocation = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'top' ? 'topleft ' : 'botright '
        let l:splitDirection = g:NERDTreeWinPos ==# 'left' || g:NERDTreeWinPos ==# 'right' ? 'vertical' : ''
        silent! execute l:splitLocation . l:splitDirection . ' new'
    else
        silent! execute 'enew'
    endif

    call creator.createWindowTree(a:dir)
    " Notice me!
    setlocal bufhidden=wipe

```

msibal6: This looks good to me!  I got these changes and renamed everything to use `Explore` instead of `Explorer` like we discussed earlier.

lib/nerdtree/creator.vim
```diff
@@ -38,6 +38,29 @@ function! s:Creator.BufNamePrefix()
     return 'NERD_tree_'
 endfunction
 
+" FUNCTION: s:Creator.CreateExploreTree(dir) {{{1
+function! s:Creator.CreateExploreTree(dir)
+    try
+        let path = g:NERDTreePath.New(a:dir)
+    catch /^NERDTree.InvalidArgumentsError/
+        call nerdtree#echo('Invalid directory name:' . a:dir)
```
rzvxa, line 46: Great work catching this, Initially I was going to suggest changing the `echo` to `echoError` but since we already use echo in the `CreateWindowTree`, It is best to leave it alone for the sake of consistency.

rzvxa (approved):

PR #1405, opened 2024-02-05
fix: session restore for nerdtree buffers.

### LINKED ISSUE #1397: Error when loading a vim session: BufLeave NERD_tree_* autocommand: Variable b:NERDTree does not exist
#### Description
Hello,

When I save and load a session with NERDTree panel open, I always get errors on session restore, from BufLeave and FocustLost autocommands, saying variable b:NERDTree does not exist. 

It feels as if the Vim session script restores the NERD_tree_1 buffer and window, the NERDTree plugin gets triggered as it sees the special buffer name, but then the buffer restored by the session script is not the same as a buffer created by NERDTree plugin, and is missing the buffer variables...

rzvxa:
### Description of Changes
Closes #1397

---
### New Version Info

This PR adds a new naming scheme for the NERDTree buffers, instead of having `NERD_tree_[num]` for all NERDTree buffers now we use `NERD_tree_tab_[num]` for tab buffers and `NERD_tree_win_[num]` for window buffers, This is then used during the restoration of a previously saved session to detect the type of given buffers and rescue them.

### DIFF
--- a/autoload/nerdtree.vim
+++ b/autoload/nerdtree.vim
@@ -234,6 +234,38 @@ function! nerdtree#pathEquals(lhs, rhs) abort
     endif
 endfunction
 
+"FUNCTION: nerdtree#onBufLeave() {{{2
+" used for handling the nerdtree BufLeave/WinLeave events.
+function! nerdtree#onBufLeave() abort
+    " detect whether we are in the middle of sourcing a session.
+    " if it is a buffer from the sourced session we need to restore it.
+    if exists('g:SessionLoad') && !exists('b:NERDTree')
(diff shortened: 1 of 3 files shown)


rzvxa:
@terminatorul This should fix the issue with restoring sessions containing NERDTree buffers. Would you give it a try?

alerque (approved):

terminatorul:
> @terminatorul This should fix the issue with restoring sessions containing NERDTree buffers. Would you give it a try?

I tried it and it works fine ! Got rid of the previous errors on session load !

PR #1419, opened 2024-06-08
compatible with incline.nvim

epheien:
Some plugins (such as incline.nvim) need autocmd event for WinEnter, BufWinEnter, etc.

So I made some slight modifications to the code to be compatible with inline, but the modifications may not be perfect. Please carefully consider when to ignore all events and when not to ignore them

### DIFF
--- a/lib/nerdtree/opener.vim
+++ b/lib/nerdtree/opener.vim
@@ -163,7 +163,7 @@ function! s:Opener._newSplit()
     " Open the new window
     try
         call nerdtree#exec('wincmd p', 1)
-        call nerdtree#exec(splitMode . ' split',1)
+        call nerdtree#exec(splitMode . ' split', 0)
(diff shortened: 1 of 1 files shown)


rzvxa:
First off thanks for this amazing work.

I'm not an `incline.nvim` user so I wasn't aware of this issue. I'll have to test it thoroughly before merging since it might have some unexpected side effects regarding how NERDTree plugins assume the NERDTree act.

It may need to be a major version bump but I hope it can be done in the next minor release(s). I have a lot of things in my backlog in order to make NERDTree play nicely with symlinks and it demands a major release since some stuff behaves drastically differently than before so if it needs a major version bump sadly it has to wait for this annoying symlink update to get ready(which has a lot of edge cases I haven't figured out yet).

Cheers mate, VimL is ugly so I love seeing community contributions(most people don't dare to touch VimL😅).

PR #1426, opened 2024-07-19
fix: typo in docs for show file lines setting

lothardp:
### Description of Changes

Just a typo fix in the docs.

The `NERDTreeShowFilesLines` setting doesn't work, and it makes sense because it's not read anywhere in the code. After some looking around I found that the right one is `NERDTreeFileLines` (as shown in the README also).

In this PR I just fix that typo in the docs so it is consistent.

---


rzvxa:
Thank you

PR #1429, opened 2024-08-24
Path.getSortOrderIndex() optimisation.

Eugenij-W:
### Description of Changes

Loading of large directories has been accelerated.


lib/nerdtree/path.vim
```diff
@@ -373,6 +374,7 @@ function! s:Path.getSortOrderIndex()
     return index(g:NERDTreeSortOrder, '*')
 endfunction
 
+
```
rzvxa, line 377:
```suggestion
```

PR #1442, opened 2025-05-08
Map multiple key sequences to the same callback

### LINKED ISSUE #1441: Map multiple key sequences to the same callback
#### Description

Can we add a "many-to-one" behavior to the key mappings, so for instance if my `.vimrc` says

        let g:NERDTreeMapHelp = ['<F1>', '?', 'h']

I would have 3 ways to open the help menu?

And in the help menu itself we would see

        " <F1>|?|h : toggle help

Thank you!

clarity20:
### Description of Changes
Closes #1441. 
The user can assign multiple key sequences to a single callback as shown in the `.vimrc` below. 
These settings are reflected in the main NERDTree panel and the detailed help panel.

NERDTree Help Menu Items
NERDTree Help Tooltip
NERDTree Settings in vimrc


rzvxa:
Hello! First off, thank you for submitting this PR.

To be honest, I'm a bit iffy about merging this. While it'd be useful and you've already taken care of the implementation, the need for multiple bindings seems a bit niche for most workflows. I personally never had the need for multiple bindings to the same functionality. Would you be able to propose a real-world scenario where this feature is required and can justify adding it to the upstream?

clarity20:
Sure! Admittedly this is not specific, but there aren't any standard key bindings across plugins for common user actions like "Help", "Delete" (both shown in the last image above), "Save", etc. and for me this idea is a nice workaround for the problem of having to use different keys in each plugin.

